PR #15, opened 2019-06-19
DCS-33 Adding elite2api dependency on docker-compose

andrewrlee:


docker-compose.yml
```diff
@@ -1,25 +1,31 @@
...
+
+  oauth-server:
+    image: mojdigitalstudio/nomis-oauth2-server:latest
     networks:
       - hmpps
+    container_name: oauth-server
+    ports:
+      - "9090:8080"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost8080/auth/health"]
```
petergphillips: missing colon

docker-compose.yml
```diff
@@ -1,25 +1,31 @@
...
     environment:
-    - PORT=3000
-    - DB_USER=use-of-force
-    - DB_PASS=use-of-force
-    - DB_SERVER=use-of-force-db
-    - DB_NAME=use-of-force
-    - DB_SSL_ENABLED=no
-    - NOMIS_AUTH_URL=http://nomis-oauth2-server:8080/auth
-    - NOMIS_AUTH_EXTERNAL_URL=http://localhost:8080/auth
+      - SERVER_PORT=8080
```
petergphillips: don't need server port

docker-compose.yml
```diff
@@ -1,25 +1,31 @@
...
+    image: mojdigitalstudio/nomis-oauth2-server:latest
     networks:
       - hmpps
+    container_name: oauth-server
+    ports:
+      - "9090:8080"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost8080/auth/health"]
+    environment:
+      - SERVER_PORT=8080
```
petergphillips: don't need server port

andrewrlee: removed

docker-compose.yml
andrewrlee: removed

docker-compose.yml
andrewrlee: fixed

PR #16, opened 2019-06-19
DCS-33 initial linking of prisoner to report

andrewrlee:
This also adds correlation id handling using the cls-hooked (in the same way as the cat tool).

We are covering the addition of an integration framework as a separate ticket.


server/config/incident.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  name: {
+  newIncident: {
```
andrewrlee: not particularly happy with this name - the first page in the workflow is 'New use of force incident' and doesn't nicely cover a unified set of info that we could describe here (which might be a bit of warning in its own right!)

Will consider this a bit further and most likely update in a follow up.

server/config.js
```diff
@@ -21,15 +21,24 @@ module.exports = {
...
-      deadline: 35000,
+  apis: {
+    oauth2: {
+      url: get('NOMIS_AUTH_URL', 'http://localhost:9090/auth', true),
+      externalUrl: get('NOMIS_AUTH_EXTERNAL_URL', get('NOMIS_AUTH_URL', 'http://localhost:9090/auth'), true),
+      timeout: {
+        response: 30000,
+        deadline: 35000,
+      },
+      apiClientId: get('API_CLIENT_ID', 'use-of-force-client', true),
```
andrewrlee: now using correct client id

server/app.js
```diff
@@ -24,7 +26,7 @@ const version = moment.now().toString()
 const production = process.env.NODE_ENV === 'production'
 const testMode = process.env.NODE_ENV === 'test'
 
-module.exports = function createApp({ signInService, formService }) {
+module.exports = function createApp({ signInService, formService, nomisService }) {
```
petergphillips: not a fan of this name - which nomis service does it correspond too?  We've renamed nomis to dps too, perhaps should be `elite2Service`.  Not precious though, would just like to be consistent

andrewrlee: fixed this: 
 * nomisClientBuilder -> elite2ClientBuilder
 * nomisService -> offenderService

server/authentication/clientCredentials.js
```diff
@@ -0,0 +1,49 @@
...
+  const correlationId = handle.get('correlationId')
+  const oauthRequest = querystring.stringify({ grant_type: 'client_credentials', username })
+
+  logger.info(
+    `Oauth request '${oauthRequest}' for client id '${config.apis.oauth2.apiClientId}' and user '${username}'`
+  )
+
+  return superagent
+    .post(oauthUrl)
+    .set('Authorization', clientToken)
```
petergphillips: since you're using superagent I think you can write:
```
    .auth(clientId, clientSecret, {type:'basic'})
``` 
and then you don't need the `generate` method

server/data/elite2ClientBuilder.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+function userGetBuilder(token) {
+  return async ({ path, query = '', headers = {}, responseType = '', raw = false } = {}) => {
+    logger.info(`Get using user credentials: calling elite2api: ${path} ${query}`)
+    try {
+      const result = await superagent
+        .get(path)
+        .query(query)
+        .set('Authorization', `Bearer ${token}`)
```
petergphillips: since you're using superagent I think you can write:
```
    .auth(token, {type:'bearer'})
```

server/data/elite2ClientBuilder.js
```diff
@@ -0,0 +1,37 @@
...
+      return userGet({ path })
+    },
+  }
+}
+function userGetBuilder(token) {
+  return async ({ path, query = '', headers = {}, responseType = '', raw = false } = {}) => {
+    logger.info(`Get using user credentials: calling elite2api: ${path} ${query}`)
+    try {
+      const result = await superagent
+        .get(path)
```
petergphillips: also be worth adding in the `keepaliveAgent` and `retries` from notm as we have a problem with requests hanging up when calling azure

petergphillips: obviously not in this commit though

PR #20, opened 2019-06-21
Drop timeouts

andrewrlee:
Dropping timeouts as these requests could now be retried twice meaning a pause of 90 seconds. This drops the time a client may have to wait, down to 30 seconds

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -26,13 +26,13 @@ module.exports = {
       url: get('NOMIS_AUTH_URL', 'http://localhost:9090/auth', true),
       externalUrl: get('NOMIS_AUTH_EXTERNAL_URL', get('NOMIS_AUTH_URL', 'http://localhost:9090/auth'), true),
       timeout: {
-        response: 30000,
-        deadline: 35000,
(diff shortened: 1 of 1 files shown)


server/config.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
       agent: {
         maxSockets: 100,
         maxFreeSockets: 10,
-        freeSocketTimeout: 30000,
+        freeSocketTimeout: 10000,
```
petergphillips: apologies, this isn't what I meant.  I meant dropping the `response` and `deadline` timeouts rather than the agent timeouts.  It is the `deadline` one that will affect retries, since if superagent doesn't get a response in that time it will retry.

petergphillips: See https://github.com/ministryofjustice/new-nomis-ui/blob/master/server/api/oauthEnabledClient.js#L58 for equivalent in new nomis.  It was slightly tricker in new nomis, since we shouldn't really retry `POST` requests as it could cause duplicate data.  I guess ideally we should have defined a timeout for get and timeout for post, but we have one at present and then divide it by three for requests that have retries...

andrewrlee: thanks, dropped the deadline timeout to 10 seconds and reverted freesockettime

PR #28, opened 2019-06-28
Alee/dcs 33 link offender

andrewrlee:
Allowing a staff member to create multiple incident reports.

These DB changes should allow the following behaviour: 
 * Each staff member can have only one in progress report per offender.
 * Staff members will not be able to view other staff members in progress reports.
 * Once a staff member marks a report as not in progress (e.g: 'complete'), then it will allow creation of a new report for that offender (by using a higher sequence number).
 * Reports are lazily created on the fly if no 'in progress' report exists for that staff member and offender.

### DIFF
--- a/migrations/20190626000000_form_db.js
+++ b/migrations/20190626000000_form_db.js
@@ -12,12 +12,12 @@ exports.up = knex =>
 
     table.string('user_id', 32).alter()
     table.string('status', 20)
-    table.unique(['booking_id', 'sequence_no'], 'booking_sequence_index')
+    table.unique(['user_id', 'booking_id', 'sequence_no'], 'user_booking_sequence_index')
   })
 
 exports.down = knex =>
   knex.schema.table('form', table => {
-    table.dropUnique(['booking_id', 'sequence_no'], 'booking_sequence_index')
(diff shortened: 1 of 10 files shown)


server/middleware/getFormData.js
```diff
@@ -1,9 +1,10 @@
 module.exports = formService => async (req, res, next) => {
   try {
-    const formData = await formService.getFormResponse(req.user.username)
+    const { bookingId } = req.params
```
andrewrlee: as we need to access request params, moving this from being router level to route level middleware

migrations/20190626000000_form_db.js
```diff
@@ -0,0 +1,25 @@
...
+      .notNullable()
+      .defaultTo(1)
+    table.bigInteger('booking_id').notNullable()
+    table
+      .timestamp('start_date')
+      .notNullable()
+      .defaultTo(knex.fn.now(6))
+
+    table.string('user_id', 32).alter()
+    table.string('status', 20)
```
andrewrlee: this could potentially be an enum but don't really know the final set of states and migrating enums is a bit fiddly in knex

PR #30, opened 2019-07-01
DCS-48 Wrapping requests in transactions.

andrewrlee:
Using cls-hooked to store a transactional db client in something akin to thread local storage.
The async middleware will set up a transaction to last the lifetime of each commit. Any thrown exception will lead to a rollback.

By transactional client will be use by default for all db requests.

This change also removes the getFormData middleware as it's simpler to not have the temporary res.locals storage and retrieval.

Fixing broken migration by providing default value for booking ids.


server/data/dataAccess/db.js
```diff
@@ -24,5 +25,25 @@ pool.on('error', error => {
 })
 
 module.exports = {
-  query: (text, params) => pool.query(text, params),
+  pool, // for testing only
+  queryWithoutTransaction: (text, params) => pool.query(text, params),
+  query: (text, params) => {
+    const ns = getNamespace('request.scope')
```
andrewrlee: could possibly look at extracting out a `requestScope` module to wrap direct access to the namespace

PR #33, opened 2019-07-01
Dcs 52 enter use of force details

GurnankCheema:
Following not yet implemented/tested
   field validation
   saving of form data to DB


integration-tests/pages/detailsPage.js
```diff
@@ -3,6 +3,20 @@ const relocationAndInjuries = require('./relocationAndInjuriesPage')
 
 export default () =>
   page('Use of force details', {
+    allRadiosYes: () => {
```
andrewrlee: might be better to just call fix `fillForm` - so we aren't leaking implementation details into the test.

package.json
```diff
@@ -5,7 +5,7 @@
   "repository": "git@github.com:ministryofjustice/use-of-force.git",
   "license": "MIT",
   "scripts": {
-    "start": "node server.js",
+    "start": "nodemon server.js",
```
andrewrlee: It would be better to just leave this as it is and use `start:dev` for development.
`start` is used in production

server/config/incident.js
```diff
@@ -8,7 +8,50 @@ module.exports = {
...
+        },
+      },
+      {
+        'personal-protection-techniques': {
+          responseType: 'requiredString',
+          validationMessage: 'Were Personal Protection techniques used?',
+        },
+      },
+      {
+        'baton-drawn-conditional': {
```
andrewrlee: think these need to be updated to match the names used in the front end

server/views/formPages/incidentMacros.njk
```diff
@@ -0,0 +1,124 @@
...
+        <div class="govuk-radios--inline" data-module="radios">
+          <div class="govuk-radios__item">
+            <input class="govuk-radios__input" id="id-{{question.primaryName}}-1" name="{{question.primaryName}}" type="radio" value="{{question.primaryOpt1}}" data-aria-controls="conditional-{{question.primaryName}}">
+            <label class="govuk-label govuk-radios__label" for='id-{{question.primaryName}}-1'>
+              {{question.primaryOpt1 | capitalize }}
+            </label>
+          </div>
+
+          <div class="govuk-radios__item">
+            <input class="govuk-radios__input" id="id-{{question.primaryName}}-2" name="{{question.primaryName}}" type="radio" value="{{question.primaryOpt2}}">
```
andrewrlee: If we could structure the question data structure in this way:
```
{
	"primaryQuestion": {
		"text": "Was control and restraint used?",
		"name": "restraint",
		"options": [{
			"value": "yes",
			"label": "Yes"
		}, {
			"value": "yes",
			"label": "Yes"
		}]
	},
	"followUpQuestion": {
		"text": "What positions were used?",
		"name": "restraintPositions",
		"options": [{
			"value": "standing",
			"label": "standing"
		}, {
			"value": "supine",
			"label": "On back (supine)"
		}, {
			"value": "prone",
			"label": "Face down (prone)"
		}, {
			"value": "kneeling",
			"label": "Kneeling"
		}]
	}
}
```

Then it would mean we wouldn't have to hard code the number of options and instead allow us to iterate the array.
See:
https://mozilla.github.io/nunjucks/templating.html#for
Also note the additional variables that are made available to the template: 
`loop.index` 
`loop.index0`

server/config/incident.js
```diff
@@ -8,7 +8,80 @@ module.exports = {
...
+        },
+      },
+      {
+        handcuffsType: {
+          responseType: 'requiredString',
+          validationMessage: 'What type of handcuffs were used?',
+        },
+      },
+    ],
     validate: false,
```
andrewrlee: shall we turn this on?

PR #38, opened 2019-07-04
DCS-54 Adding tasklist page and fixing not saving initial form

andrewrlee:
The tasklist page will be the first page that a staff member will visit
after clicking the use-of-force link.

It will give context about what the process is and how much of the form is
complete.

Screenshot 2019-07-04 at 10 06 09

Also fixing a bug where we ignored the first lot of form data if the form
didn't previously exist.


server/views/pages/tasklist.html
```diff
@@ -0,0 +1,48 @@
...
+              <a href="/form/incident/relocationAndInjuries/{{bookingId}}" data-qa-relocation-and-injuries-link>Relocation and Injuries</a>
+              {% if form.incident.relocationAndInjuries %}
+              <strong class="govuk-tag app-task-list__task-completed" data-qa-relocation-and-injuries-completed>Completed</strong>
+              {% endif %}
+            </li>
+            <li class="app-task-list__item">
+                <a href="/form/incident/evidence/{{bookingId}}" data-qa-evidence-link>Evidence</a>
+                {% if form.incident.evidence %}
+                <strong class="govuk-tag app-task-list__task-completed" data-qa-evidence-completed>Completed</strong>
+                {% endif %}
```
Nimphal: Not a blocker by any means, but I am wondering if the `<strong>Completed</strong>` could be made a partial? 

Something like:
```
<%- include('taksCompleted', {task: 'evidence'}); %>
```

And then the partial looks like:
```
<strong class="govuk-tag app-task-list__task-completed" data-qa-{% task %}-completed>Completed</strong>
```

I haven't tested that the syntax works, so it probably doesn't.

andrewrlee: I could maybe look at doing each entire row as a partial in a follow up? 

(This was based on the original form builder tasklist code)

PR #39, opened 2019-07-04
DT-72: Added retries to health check

SimonMitchellMOJ:


server/data/healthcheck.js
```diff
@@ -1,41 +1,57 @@
...
+  maxFreeSockets: config.apis.elite2.agent.maxFreeSockets,
+  freeSocketTimeout: config.apis.elite2.agent.freeSocketTimeout,
+}
 
-          if (result.status === 200) {
-            return resolve('OK')
-          }
+function serviceCheckFactory(name, url) {
+  const keepaliveAgent = url.startsWith('https') ? new HttpsAgent(agentOptions) : new Agent(agentOptions)
+  logger.warn(`We have constructed a new service check here for ${url}`)
```
andrewrlee: meh.. was this just for debug during development? do we need to log this

server/data/healthcheck.js
```diff
@@ -1,41 +1,57 @@
...
+          try {
+            if (error) {
+              logger.error(error.stack, `Error calling ${name}`)
+              reject(error)
+            } else if (result.status === 200) {
+              resolve('OK')
+            } else {
+              reject(result.status)
+            }
+          } catch (apiError) {
```
andrewrlee: not quite sure what would throw this error? the code in the try seems to only log or return a promise?

andrewrlee: when would this be thrown?

server/data/healthcheck.js
SimonMitchellMOJ: Removed

server/data/healthcheck.js
SimonMitchellMOJ: code removed

PR #43, opened 2019-07-08
DCS-37  initial details page

andrewrlee:
This is still a bit of a WIP.

Screenshot 2019-07-08 at 13 22 30

Still need to:
* Add incident date
* Convert use of force type to a question
* Sanitise incoming data - remove empty items from array, convert ids to numbers
* Workout what we will do regarding search for staff in the MVP
* Look at bundling js

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,7 @@
     "joi": "^14.3.1",
     "joi-date-extensions": "^1.2.0",
     "joi-postcode": "^3.0.0",
+    "jquery": "^3.4.1",
     "jwt-decode": "^2.2.0",
     "knex": "^0.16.5",
     "moment": "^2.24.0",
--- a/server/config/incident.js
+++ b/server/config/incident.js
@@ -1,6 +1,25 @@
 module.exports = {
   newIncident: {
-    fields: [],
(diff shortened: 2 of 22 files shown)


server/config/incident.js
```diff
@@ -1,6 +1,25 @@
...
+      {
+        locationId: {
+          responseType: 'requiredNumber',
+          validationMessage: 'Where did the incident occur?',
+        },
+      },
+      {
+        forceType: {
+          responseType: 'requiredString',
+          validationMessage: 'Where did the incident occur?',
```
petergphillips: same message for both?

andrewrlee: fixed - we haven't enabled validation yet (it's to be covered in a later story and we are still awaiting real content)

package.json
```diff
@@ -109,6 +109,7 @@
     "joi": "^14.3.1",
     "joi-date-extensions": "^1.2.0",
     "joi-postcode": "^3.0.0",
+    "jquery": "^3.4.1",
```
petergphillips: Might be better to use a CDN instead - https://jquery.com/download/

PR #48, opened 2019-07-12
DCS-56 fixing issue where radio buttons were being assigned incorrect…

andrewrlee:
… name

### DIFF
--- a/assets/js/add-another.js
+++ b/assets/js/add-another.js
@@ -53,7 +53,7 @@ AddAnother.prototype.updateAttributes = function(index, item) {
     el.id = $(el)
       .attr('data-id')
       .replace(/%index%/, index)
-    ;($(el).prev('label')[0] || $(el).parents('label')[0]).htmlFor = el.id
+    ;($(el).prev('label')[0] || $(el).parents('label')[0] || $(el).next('label')[0]).htmlFor = el.id
   })
 }
 
--- a/server/views/formPages/incident/relocationAndInjuries.html
+++ b/server/views/formPages/incident/relocationAndInjuries.html
@@ -288,24 +288,32 @@
                   }}
                   {{ govukRadios({
                     classes: "govuk-radios--inline",
-                    idPrefix: 'staff-member-went-to-hospital[0][name]',
                     name: 'staffMemberWentToHospital[0][name]',
                     fieldset: {
                       legend: {
                         text: "Did they go to hospital?"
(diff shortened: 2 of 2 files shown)


assets/js/add-another.js
```diff
@@ -53,7 +53,7 @@ AddAnother.prototype.updateAttributes = function(index, item) {
     el.id = $(el)
       .attr('data-id')
       .replace(/%index%/, index)
-    ;($(el).prev('label')[0] || $(el).parents('label')[0]).htmlFor = el.id
+    ;($(el).prev('label')[0] || $(el).parents('label')[0] || $(el).next('label')[0]).htmlFor = el.id
```
andrewrlee: the HMCTS code only coped when the element was preceded or surrounded by a label.
Radio buttons have labels afterwards

server/views/formPages/incident/relocationAndInjuries.html
```diff
@@ -288,24 +288,32 @@ <h2 class="govuk-heading-l govuk-!-margin-top-6">Injuries sustained and healthca
                   }}
                   {{ govukRadios({
                     classes: "govuk-radios--inline",
-                    idPrefix: 'staff-member-went-to-hospital[0][name]',
```
andrewrlee: got rid of this - as it's a bit more readable to know what ids were being used

server/views/formPages/incident/relocationAndInjuries.html
```diff
@@ -288,24 +288,32 @@ <h2 class="govuk-heading-l govuk-!-margin-top-6">Injuries sustained and healthca
...
                     items: [
                       {
                         value: "yes",
                         text: "Yes",
-                        checked: data.staffMemberWentToHospital[0].name === 'yes'
+                        id: 'staff-member-went-to-hospital[0]-yes',
+                        checked: data.staffMemberWentToHospital[0].name === 'yes',
+                        attributes: {
+                          'data-name': 'staffMemberWentToHospital[%index%][name]',
+                          'data-id': 'staff-member-went-to-hospital[%index%]-yes'
```
andrewrlee: these attributes are required - after the clone, the js will find the attributes and do a find and replace on the name and id using these templates.

andrewrlee: might need to do something similar in the case where there are some items already persisted.

PR #51, opened 2019-07-16
DCS 61: User can now complete Evidence page.

GurnankCheema:
http://localhost:3000/form/incident/evidence/-35

First commit following creation of components in the Evidence page.
This comprises 4 questions.

Also completed bugfix: empty fields after doing Add Another no longer persisted in DB


server/config/incident.js
```diff
@@ -164,14 +164,13 @@ module.exports = {
       },
       {
         staffMemberNeedingMedicalAttention: {
-          responseType: 'requiredString',
-          validationMessage: 'What is the name of the staff who needed medical attention?',
+          sanitiser: (vals = []) => vals.reduce((res, val) => (val.name && val.name.trim() ? [...res, val] : res), []),
```
timharrison-moj: Could this be clearer? I am not sure that I understand what it is doing.

server/views/formPages/incident/evidence.html
```diff
@@ -1,5 +1,337 @@
...
+<script src="/assets/add-another.js"></script>
+<script>
+new AddAnother($('.add-another-tag'))
+new AddAnother($('.add-another-camera'))
+
+</script>
+
+
+   
+
```
timharrison-moj: Can you think of any way to break down this template into smaller, perhaps reusable, units? I am sure it does the job, but it seems quite long. Perhaps a per-question template?

timharrison-moj: Might be worth thinking about ways to pull in the data values rather than embedding in the HTML?

GurnankCheema: Yes, it is rather long. But I did it this way to stay consistent with other similar pages in the application. The other thing is, although the components do look the same, there are differences in the way they are laid out. For example, some are 2 radios horizontally, others are 3 vertically and then some expose hidden panels that display different numbers of input fields when clicking the 'yes' radio button.

server/config/incident.js
GurnankCheema: Good idea. I'll have a go.

GurnankCheema: Have refactored and created a new function called inputsExcludingEmptyStrings()

timharrison-moj: Reads better!

PR #59, opened 2019-07-24
DCS-64 Storing the offender number and reporter name

andrewrlee:


migrations/20190724000000_form_db.js
```diff
@@ -0,0 +1,11 @@
+exports.up = knex =>
+  knex.schema.alterTable('form', table => {
+    table.string('offender_no', 32).notNullable()
+    table.string('reporter_name', 128).notNullable()
```
andrewrlee: (I'm planning on clearing out the DB in dev as there is no sensible offender number to pre populate existing rows with)

PR #61, opened 2019-07-25
Dcs 47 check answers before submission

GurnankCheema:
Display of user inputs in the Check Answers page
Cypress tests to verify correct details are being displayed
Capitalised radio button values so they are as required in the display.


integration-tests/integration/createIncident/incident-creation.spec.js
```diff
@@ -16,13 +17,15 @@ context('Logging in', () => {
...
     const newIncidentPage = tasklistPage.startNewForm()
+    newIncidentPage.fillForm()
     const detailsPage = newIncidentPage.save()
     detailsPage.fillForm()
     const relocationAndInjuriesPage = detailsPage.save()
     relocationAndInjuriesPage.fillForm()
     const evidencePage = relocationAndInjuriesPage.save()
     evidencePage.fillForm()
-    evidencePage.save()
+    const checkInputs = evidencePage.save()
```
andrewrlee: maybe call this 'checkAnswersPage' - this is what we refer to it in other places

integration-tests/pages/checkAnswersPage.js
```diff
@@ -4,6 +4,44 @@ const clickSubmit = () => cy.get('[data-submit]').click()
 
 export default () =>
   page('Check your answers before submitting the report', {
+    verifyInputs: () => {
+      cy.get('.govuk-grid-column-full > :nth-child(2) > :nth-child(2) > :nth-child(2)').contains('Norman Smith')
```
andrewrlee: As discussed - can move this to use data-qa attributes in a follow up, when you render date too
e.g: `data-qa='name'` ?

server/routes/checkAnswers.js
```diff
@@ -3,7 +3,8 @@ const bodyParser = require('body-parser')
 const flash = require('connect-flash')
 const asyncMiddleware = require('../middleware/asyncMiddleware')
 
-module.exports = function Index({ formService, authenticationMiddleware }) {
+// eslint-disable-next-line
```
andrewrlee: the eslint-disable warnings should only be used very sparingly - can you fix the issue which has lead you to put this in?

server/routes/checkAnswers.js
```diff
@@ -21,7 +22,109 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
...
+      // retrieve form inputs
+      // eslint-disable-next-line
+      const loadForm = async req => {
+        const { form_response: formObject = {}, id: formId } = await formService.getFormResponse(
+          req.user.username,
+          bookingId
+        )
+        return { formId, formObject }
+      }
+      const { formObject } = await loadForm(req, res)
```
andrewrlee: you could merge this method and the call:
 
```
      const { form_response: formObject = {} } = await formService.getFormResponse(
          req.user.username,
          bookingId
        )
```

server/routes/checkAnswers.js
```diff
@@ -47,3 +150,64 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
...
+  return allElements
+}
+
+// utility function - convert to string when input WILL be nested objects
+const convertArrayOfObjectsToString = (dataArray = []) => {
+  let finalString = ''
+  dataArray.forEach(element => {
+    finalString += `${element.name}, `
+  })
+  return finalString.slice(0, -2)
```
andrewrlee: think this could use array.join? and remove this method: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

server/routes/checkAnswers.js
```diff
@@ -47,3 +150,64 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
...
+}
+
+// titlecase all person names
+const toTitleCase = (str = '') => {
```
andrewrlee: we have a method in utils that we could use instead of this: `properCaseName`

server/routes/checkAnswers.js
```diff
@@ -47,3 +150,64 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
...
+    return 'none'
+  }
+  return tagsAndEvidence
+    .map(item => {
+      return `${item.name} ${item.description}`
+    })
+    .join(`<br/>`)
+}
+
+// how many officers involved
```
andrewrlee: the comments on `howManyOfficersInvolved` and `typeOfHandcuffsUsed` seem a bit superfluous - the method name should give enough context

server/views/formPages/incident/details.html
```diff
@@ -120,11 +120,11 @@ <h1 class="govuk-heading-xl mainHeading">Use of force details</h1>
               "name": "guidingHold",
               "value": data.guidingHold,
               "options": [{
-                "value": "yes",
+                "value": "Yes",
```
andrewrlee: As discussed - we should probably ticket this up to store this more as a "code" rather than a "label"

server/routes/checkAnswers.js
```diff
@@ -47,3 +150,64 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
 
   return router
 }
+
+// utility function - convert to string where input MAY be an array
+
+const convertNestedArraysToString = allElements => {
```
andrewrlee: Rather than creating a generic method for this - it might be worth making it very specific: 
e.g: `const getRestraintPositions = ({ restraintPositions }) => ...`

It's probably only worth trying to make a method generic if you have a few places where it's being used - this seems very specific to this one use case. 

(If you did need to create a generic method, then it should probably live with other more generic code (utils?) and be tested independently)

server/routes/checkAnswers.js
```diff
@@ -21,7 +22,109 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
...
+        'Offender number': offenderDetail.offenderNo,
+        Location: offenderLocation.description,
+        'Use of force planned': viewObject.newIncident.forceType,
+        'Staff involved': toTitleCase(convertArrayOfObjectsToString(viewObject.newIncident.involved)),
+        Witnesses: toTitleCase(convertArrayOfObjectsToString(viewObject.newIncident.witnesses)),
+      }
+      viewObject.newIncident = newIncident
+      viewObject.offenderDetail = offenderDetail
+
+      // details object. Initilise to previous values or empty object to prevent 'undefined' errors
```
andrewrlee: As discussed it would be worth breaking up this large method and extracting each section into separate simple functions, so you end up with no mutation, maybe something like this: 
```
const viewModel = {
   details: buildDetails(),
  relocationAndInjuries: buildRelocationAndInjuries(),
  evidence: buildEvidence(),
  // ... etc..
}
```

integration-tests/integration/createIncident/incident-creation.spec.js
GurnankCheema: I had called it checkInputs rather than checkAnswersPage because checkAnswersPage is used as a const a few lines below. 
So, Have done as above and removed the const from line 33 so that checkAnswersPage can be reassigned thus: checkAnswersPage = tasklistPageAfterAllPartsComplete.goToAnswerPage()

integration-tests/pages/checkAnswersPage.js
GurnankCheema: ok

server/routes/checkAnswers.js
GurnankCheema: it was a line I had forgotten to take out. I had initially included the lint disable  because I was erroneously 'requiring' offenderService in line 5 as well as passing it in as an arg in line 7. 
However have removed the 'require' so ok to take out the lint disable

server/routes/checkAnswers.js
GurnankCheema: ok

server/routes/checkAnswers.js
GurnankCheema: agreed. It is much more succinct. 
replaced with 
const convertArrayOfObjectsToString = (dataArray = []) => {
  return dataArray.map(element => element.name).join(', ')
}

GurnankCheema: agreed. It is much more succinct.
replaced with
const convertArrayOfObjectsToString = (dataArray = []) => {
return dataArray.map(element => element.name).join(', ')
}

server/routes/checkAnswers.js
GurnankCheema: agreed and have replaced occurrences of toTitleCase with properCaseName

server/routes/checkAnswers.js
GurnankCheema: Done. It's just a personal preference to add comments as I find it quicker to distinguish code sections on a page

server/views/formPages/incident/details.html
GurnankCheema: I will do this as a separate ticket because it will impact across every page in the application.

server/routes/checkAnswers.js
GurnankCheema: I gave it a generic name on the chance that I may be able to use it elsewhere in the application. In the event, it hasn't been needed elsewhere. So yes, I agree that it can have a specific name.
Renamed and refactored.

server/routes/checkAnswers.js
GurnankCheema: will do this as a separate ticket together with renaming Yes to yes etc

server/routes/checkAnswers.js
```diff
@@ -47,3 +140,47 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
...
 }
+
+const getRestraintPositions = positions => {
+  if (Array.isArray(positions)) {
+    return positions.join(', ')
+  }
+  return ''
+}
+
+const convertArrayOfObjectsToString = (dataArray = []) => {
```
andrewrlee: maybe rename this to `extractNames` ?

server/routes/checkAnswers.js
andrewrlee: fair enough, that has actually stung me a couple of times.
I think this could be moved in to utils and tested independently - but could do this in a follow up?

server/routes/checkAnswers.js
GurnankCheema: I've used that function for 'Body worn cameras' too so I'm not sure about changing it to something related just to names?

andrewrlee: it extracts a field called name, so that suggests that the data is being incorrectly stored.

server/routes/checkAnswers.js
GurnankCheema: I've refactored the function into smaller units that each build up the newIncident, details, relocation and evidence objects. 
You may be wondering why I've needed lines 31 to 34 of checkAnswers.js but I could only get the functionality to work by including that code.

server/routes/checkAnswers.js
GurnankCheema: I thought extract() was not descriptive enough so have renamed to convertArrayOfObjectsToStringUsingSpecifiedKey() instead. 
The refactored function now takes another arg that is the end of the name property for the element, i.e. cameraNum, name.

server/config/incident.js
```diff
@@ -237,10 +232,10 @@ module.exports = {
   },
 }
 
-function inputsExcludingEmptyStrings(inputs) {
+function inputsExcludingEmptyStrings(attr = 'name', inputs = []) {
   const sanitisedArray = []
```
steverendell: would a filter and util be nicer here, e.g.
const sanitisedArray = inputs.filter(input => !isBlank(input))

server/routes/checkAnswers.js
```diff
@@ -47,3 +65,110 @@ module.exports = function Index({ formService, authenticationMiddleware }) {
...
+            convertArrayOfObjectsToStringUsingSpecifiedKey('name', relocationAndInjuries.staffNeedingMedicalAttention)
+          )}`
+        : ''),
+    'Staff taken to hospital': staffTakenToHospital(relocationAndInjuries.staffNeedingMedicalAttention),
+  }
+}
+
+const createEvidenceObj = (evidence = {}) => {
+  return {
+    'Evidence bagged and tagged': baggedAndTaggedEvidence(evidence.evidenceTagAndDescription, evidence.baggedEvidence),
```
steverendell: Typo

server/views/formPages/incident/evidence.html
```diff
@@ -61,34 +58,29 @@ <h1 class="govuk-heading-xl mainHeading">Evidence</h1>
...
+                        'data-name': 'evidenceTagAndDescription[%index%][evidenceTagReference]',
+                        'data-id': 'evidence-tag-and-description[%index%][evidenceTagReference]'
                       }
                     })
                   }} 
-
-                  {% set index = loop.index0 %}
-
-                  <!--this assignment is required because can't access both tagNumber and evidenceDescriptions in same loop.
-                  The value for index will change on each iteration so will keep both data keys in sync -->
```
steverendell: General comment about text areas: the cat tool uses govukCharacterCount which gives the user some guidance on whether there is too much input text

server/routes/checkAnswers.js
GurnankCheema: good spot!

server/views/formPages/incident/evidence.html
GurnankCheema: We did discuss character count with the Andy Hudson, the designer. He is already looking at what the limit should be. Will implement when he responds.

server/config/incident.js
GurnankCheema: done but have changed function name to removeEmptyValues

PR #62, opened 2019-07-25
DCS-47 Fixing integration tests - make code more resilient when no lo…

andrewrlee:
…cation id has been specified


server/services/offenderService.js
```diff
@@ -50,6 +50,9 @@ module.exports = function createOffendersService(elite2ClientBuilder) {
   }
 
   const getLocation = (token, locationId) => {
+    if (!locationId) {
```
andrewrlee: the main issue was this - it was always attempting to load a location - even if one hadn't been saved previously.

PR #63, opened 2019-07-26
DCS-60 Fixing redirect loop & adding token refreshing

andrewrlee:
Token refresh was not implemented - but there was some code which attempted to redirect to login 5 minutes before the token expired (Not quite sure why).

The redirect would always be intercepted by the same code, so an infinite redirect loop would occur.

I've fixed this by back porting token refreshing functionality from categorisation tool.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -170,13 +170,26 @@ module.exports = function createApp({ signInService, formService, offenderServic
     app.use(csurf())
   }
 
-  // token refresh
+  // JWT token refresh
   app.use(async (req, res, next) => {
-    if (req.user) {
+    if (req.user && req.originalUrl !== '/logout') {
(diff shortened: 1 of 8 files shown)


server/app.js
```diff
@@ -170,13 +170,26 @@ module.exports = function createApp({ signInService, formService, offenderServic
     app.use(csurf())
   }
 
-  // token refresh
+  // JWT token refresh
   app.use(async (req, res, next) => {
-    if (req.user) {
+    if (req.user && req.originalUrl !== '/logout') {
```
andrewrlee: don't want to refresh on `/logout`- otherwise this leads to a redirect loop in the case of a token refresh error being thrown below.

PR #65, opened 2019-07-30
DCS-64 Refactoring incident persistence

andrewrlee:
Several changes to the data layer: 
* Renaming form service to incident service (and renaming associated tables)
* renaming incident start_date to created_date
* Storing incident date separately from json payload
* Storing staff involved in an incident in separate table

### DIFF
--- a/server/config/incident.js
+++ b/server/config/incident.js
@@ -1,5 +1,6 @@
 const moment = require('moment')
 const { isBlank } = require('../utils/utils')
+const { EXTRACTED } = require('./fieldType')
 
 module.exports = {
   newIncident: {
@@ -8,6 +9,7 @@ module.exports = {
         incidentDate: {
           responseType: 'requiredString',
           sanitiser: val => moment(val, 'DD/MM/YYYY-HH:mm').toDate(),
+          fieldType: EXTRACTED,
         },
       },
       {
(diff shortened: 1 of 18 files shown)


server/data/formClient.js
```diff
@@ -1,54 +1,85 @@
...
+}
+
+const deleteInvolvedStaff = incidentId => {
+  return db.query({
+    text: `delete from involved_staff where incident_id = $1`,
+    values: [incidentId],
+  })
+}
+
+const insertInvolvedStaff = async (incidentId, staff) => {
```
andrewrlee: using `pg_format` to safely allow bulk insert without sql injection

server/config/incident.js
```diff
@@ -32,6 +34,7 @@ module.exports = {
       {
         involved: {
           sanitiser: vals => removeEmptyValues('name', vals),
+          fieldType: EXTRACTED,
```
andrewrlee: Marking these extracted means that they won't be included as part of the json payload and can be intercepted later:

before:
```
{
	"incidentDate": "",
	"name": "",
	"involved": "",
	"evidence": ""
}
`` `
now: 
```{
	"incidentDate": "",
	"involved": "",
	"payload": {
		"name": "",
		"evidence": ""
	}
}
```

PR #68, opened 2019-07-30
DCS-68: HTML structure for Check Answers page corrected

GurnankCheema:
http://localhost:3000/check-answers/-1

The structure was taken from the GDS components guide but hadn't been applied correctly.
Whilst this didn't impact on the data being displayed, it did mean some format couldn't be applied.
Structure updated and also new Nunjucks macro created enabling similar sections to be looped through rather than being coded multiple times.
Cypress test updated although this may change again in a future ticket.


integration-tests/pages/checkAnswersPage.js
```diff
@@ -5,45 +5,39 @@ const clickSubmit = () => cy.get('[data-submit]').click()
 export default () =>
   page('Check your answers before submitting the report', {
     verifyInputs: () => {
-      cy.get('.govuk-grid-column-full > :nth-child(2) > :nth-child(2) > :nth-child(2)').contains('Norman Smith')
-      cy.get('.govuk-grid-column-full > :nth-child(2) > :nth-child(3) > :nth-child(2)').contains('A1234AC')
```
andrewrlee: we still have an outstanding task to move these to use data-qa tags, so will be done as a follow up

PR #71, opened 2019-07-31
DCS-56 and 61 Fixed HTML validation errors

GurnankCheema:
http://localhost:3000/form/incident/relocationAndInjuries/-1
http://localhost:3000/form/incident/evidence/-1

Ran the Relocation and Injuries and Evidence pages through HTMLM validator. It revealed unnecessary definition of rows and columns for textareas, missing divs and missing id attribute. Corrected erorrs and rechecked. Updated Cypress tests.

### DIFF
--- a/server/routes/checkAnswers.js
+++ b/server/routes/checkAnswers.js
@@ -75,6 +75,9 @@ const getRestraintPositions = positions => {
   if (Array.isArray(positions)) {
     return positions.join(', ')
   }
+  if (positions) {
+    return positions
+  }
   return ''
 }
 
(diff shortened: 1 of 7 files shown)


server/routes/checkAnswers.js
```diff
@@ -128,37 +131,38 @@ const createDetailsObj = (details = {}) => {
...
-      details.restraint +
-      (details.restraint === 'Yes' ? ` - ${getRestraintPositions(details.restraintPositions)}` : ''),
-    'Handcuffs used':
-      details.handcuffsApplied + (details.handcuffsApplied === 'Yes' ? typeOfHandcuffsUsed(details.handcuffsType) : ''),
+    'Baton drawn': whenPresent(details.batonDrawn, value =>
+      value === 'Yes' ? wasWeaponUsed(details.batonUsed) : 'No'
+    ),
+    'PAVA drawn': whenPresent(details.pavaDrawn, value => (value === 'Yes' ? wasWeaponUsed(details.pavaUsed) : 'No')),
+    'Guiding hold used': whenPresent(details.guidingHold, value =>
+      value === 'Yes' ? `Yes ${howManyOfficersInvolved(details.guidingHoldOfficersInvolved)}` : 'No'
```
andrewrlee: meh.. these methods are a bit inconsistent as to whether they will have `Yes` or `Yes - ` or `` as part of the prefix. 
Kind of prefer the more explicit approach of moving `Yes -` up to this level, like the `getRestraintPositions` example but not precious.

server/views/pages/pagesMacros.njk
```diff
@@ -16,9 +14,9 @@
       <div class="govuk-summary-list__row">
             <dt class="govuk-summary-list__key"> {{key}} </dt>
             <dd class="govuk-summary-list__value"> 
-              {% if not value or regExp.test(value) %} 
+              {% if not value or value.length  == 0 %} 
                 &ndash;  
-              {% elif key === "Evidence bagged and tagged" %} 
+              {% elif key === "Evidence bagged and tagged" and value !== "No" %}
```
andrewrlee: guess - this won't be necessary once we fix the config. We should make a note to remove this then...

PR #77, opened 2019-08-02
DCS-67 Updating content based on designer feedback

andrewrlee:


server/views/formPages/incident/newIncident.html
```diff
@@ -51,7 +51,7 @@ <h1 class="govuk-heading-xl mainHeading">New use of force incident</h1>
         name: "forceType",
         fieldset: {
           legend: {
-            text: "Was the use of force plannned?",
+            text: "Was the use of force planned?",
             isPageHeading: false,
```
Nimphal: 😆

andrewrlee: That's been in for a such a long time and noooone noticed!

PR #79, opened 2019-08-02
DCS-74: Date and Time of incident displayed

GurnankCheema:
http://localhost:3000/check-answers/-1

The Check Answers page now displays 2 new rows for date and time.
Cypress tests updated

### DIFF
--- a/integration-tests/pages/checkAnswersPage.js
+++ b/integration-tests/pages/checkAnswersPage.js
@@ -1,12 +1,20 @@
 const page = require('./page')
 
 const clickSubmit = () => cy.get('[data-submit]').click()
+const incidentDateFormat = new RegExp(Cypress.moment().format('DD/MM/YYYY'))
+const incidentTimeFormat = new RegExp(Cypress.moment().format('HH:mm'))
 
 export default () =>
   page('Check your answers before sending the report', {
     verifyInputs: () => {
       cy.get('[data-qa="offenderName"]').contains('Norman Smith')
       cy.get('[data-qa="offenderNumber"]').contains('A1234AC')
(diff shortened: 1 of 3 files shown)


integration-tests/pages/checkAnswersPage.js
```diff
@@ -1,12 +1,20 @@
 const page = require('./page')
 
 const clickSubmit = () => cy.get('[data-submit]').click()
+const incidentDateFormat = new RegExp(Cypress.moment().format('DD/MM/YYYY'))
+const incidentTimeFormat = new RegExp(Cypress.moment().format('HH:mm'))
```
andrewrlee: won't this fail if the start of the test and end of the test span the minute boundary

server/routes/checkAnswers.js
```diff
@@ -110,14 +118,16 @@ const typeOfHandcuffsUsed = handcuffsType => {
...
+const createNewIncidentObj = (offenderDetail, description, newIncident = {}, involvedStaff, incidentDate) => {
   return {
     offenderName: offenderDetail.displayName,
     offenderNumber: offenderDetail.offenderNo,
     location: description,
     forceType: newIncident.forceType,
     staffInvolved: convertArrayOfObjectsToStringUsingSpecifiedKey('name', involvedStaff),
     witnesses: convertArrayOfObjectsToStringUsingSpecifiedKey('name', newIncident.witnesses),
+    incidentDate: moment(incidentDate).format('DD/MM/YYYY'),
+    incidentTime: moment(incidentDate).format('HH:mm'),
```
andrewrlee: this could just be a single variable - the reason it's two on the other page is that we need to separate it for editing - as each part is input separately

PR #82, opened 2019-08-05
DCS-70-upgrade-govuk-frontend-tollkit-to-3.0.0

GurnankCheema:
Application upgraded from govuk-frontend toolkit v2.12.0 to 3.0.0

Required paths to be suffixed with /govuk and for data-module attributes to be prefixed with govuk. See https://github.com/alphagov/govuk-frontend/releases

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -124,7 +124,7 @@ module.exports = function createApp({ signInService, incidentService, offenderSe
     '../assets',
     '../assets/stylesheets',
     '../assets/js',
-    '../node_modules/govuk-frontend/assets',
+    '../node_modules/govuk-frontend/govuk/assets',
     '../node_modules/govuk-frontend',
   ].forEach(dir => {
     app.use('/assets', express.static(path.join(__dirname, dir), cacheControl))
--- a/server/utils/nunjucksSetup.js
+++ b/server/utils/nunjucksSetup.js
@@ -1,17 +1,10 @@
 const nunjucks = require('nunjucks')
 
 module.exports = (app, path) => {
-  const njkEnv = nunjucks.configure(
-    [
-      path.join(__dirname, '../../server/views'),
-      'node_modules/govuk-frontend/',
-      'node_modules/govuk-frontend/components/',
-    ],
-    {
-      autoescape: true,
-      express: app,
-    }
-  )
+  const njkEnv = nunjucks.configure([path.join(__dirname, '../../server/views'), 'node_modules/govuk-frontend/'], {
+    autoescape: true,
+    express: app,
+  })
 
   njkEnv.addFilter('findError', (array, formFieldId) => {
     const item = array.find(error => error.href === `#${formFieldId}`)
--- a/server/views/formPages/incident/evidence.html
+++ b/server/views/formPages/incident/evidence.html
@@ -1,12 +1,12 @@
 {% extends "../formTemplate.html" %}
-{% from "input/macro.njk" import govukInput %}
-{% from "radios/macro.njk" import govukRadios %}
-{% from "select/macro.njk" import govukSelect %}
-{% from "checkboxes/macro.njk" import govukCheckboxes %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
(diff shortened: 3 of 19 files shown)


server/utils/nunjucksSetup.js
```diff
@@ -4,8 +4,8 @@ module.exports = (app, path) => {
   const njkEnv = nunjucks.configure(
     [
       path.join(__dirname, '../../server/views'),
-      'node_modules/govuk-frontend/',
-      'node_modules/govuk-frontend/components/',
+      'node_modules/govuk-frontend/govuk/',
```
petergphillips: is this correct?
https://github.com/alphagov/govuk-frontend/releases states
>Change the list of paths in nunjucks.configure so that the only GOV.UK Frontend path is node_modules/govuk-frontend/:

server/views/formPages/incident/evidence.html
```diff
@@ -21,7 +21,7 @@ <h1 class="govuk-heading-xl mainHeading">Evidence</h1>
           Was any evidence bagged and tagged?
         </legend>
 
-        <div class="govuk-radios" data-module="radios">
+        <div class="govuk-radios" data-module="govuk-radios">
```
petergphillips: not sure why this is required.
From https://github.com/alphagov/govuk-frontend/releases
>Update and add data-module attributes
>You do not need to do anything if you're using Nunjucks macros and the initAll function.
>If you are not using Nunjucks macros, add a govuk- prefix to data-module attribute values

Does that suggest we're not using `initAll` correctly?

petergphillips:
Also from release notes:
>2. If you've extended the page template, add govuk/ to the template path:
>{% extends "govuk/template.njk" %}
>
>Change the import paths in your components so they include govuk/components/. For example:
>{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
seems to be missing from changes

server/views/formPages/incidentMacros.njk
```diff
@@ -99,7 +99,7 @@
           {% endfor %}
 
           {# followup #}
-          <div class="govuk-checkboxes__conditional clear-both" id="conditional-{{name}}">
+          <div class="govuk-radios__conditional clear-both" data-module="govuk-checkboxes" id="conditional-{{name}}">
```
petergphillips: this doesn't look right - why use a `radios` class for `checkboxes`?

server/app.js
```diff
@@ -112,7 +112,7 @@ module.exports = function createApp({ signInService, incidentService, offenderSe
         debug: true,
         outputStyle: 'compressed',
         prefix: '/stylesheets/',
-        includePaths: ['node_modules/govuk-frontend'],
+        includePaths: ['node_modules/govuk-frontend/govuk'],
```
petergphillips: not sure about this.  They've gone out of their way to ensure that they prefix everything with `govuk` to ensure that it doesn't conflict and we're then ignoring it.  Their suggested import is
```@import "govuk/all";
``` 
so that `govuk` still there

petergphillips: The same applies with `all.js` that we are merging into one assets directory first.

server/utils/nunjucksSetup.js
GurnankCheema: As discussed, have changed to node_modules/govuk-frontend/

server/views/formPages/incident/evidence.html
GurnankCheema: We are using a mixture of HTML and Nunjucks because the design requested customised components. Because of this mix, I still had to isert the govuk- prefix

server/views/formPages/incidentMacros.njk
GurnankCheema: Changed from checkboxes to radios, even though component is in fact a checkbox, because the component didn't function with the new version of the toolkit. 
This is a non-standard Govuk component

server/app.js
GurnankCheema: agreed and updated

PR #88, opened 2019-08-09
DCS-86 adding ability to link errors to individual array items

andrewrlee:
Adding ability to link errors to individual array items when multiple values can be input via add-another component. 

Also allowing focusing on the first field of the first item when there is something wrong with the overall structure of an array. 

Also adding validation tests

We currently display the same message for each invalid item: 
<kbd>
Screenshot 2019-08-09 at 15 48 46
</kbd>

So although the links now work and point to the correct item - we will need some content updating (and a way to get access to the index which might be a bit gnarly...)

### DIFF
--- a/server/utils/fieldValidation.js
+++ b/server/utils/fieldValidation.js
@@ -77,6 +77,20 @@ const fieldOptions = {
     }),
 }
 
+const getHref = (fieldConfig, error) => {
+  const [head, ...sections] = error.path
+  const { firstFieldName } = fieldConfig[head]
+
+  if (sections.length === 0 && firstFieldName) {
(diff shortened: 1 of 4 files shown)


server/utils/fieldValidation.js
```diff
@@ -77,6 +77,20 @@ const fieldOptions = {
     }),
 }
 
+const getHref = (fieldConfig, error) => {
+  const [head, ...sections] = error.path
+  const { firstFieldName } = fieldConfig[head]
+
+  if (sections.length === 0 && firstFieldName) {
```
JamieDraperUK: You could get away with sections.length && if you wanted.

PR #119, opened 2019-08-22
DCS-98 Fixing issue with single checkbox selection

andrewrlee:
Fixing issue where, if only a single item was selected, it would lead to a single value being passed rather than an array. 

Also improving tests on use-of-force-details page

### DIFF
--- a/integration-tests/db/db.js
+++ b/integration-tests/db/db.js
@@ -22,13 +22,14 @@ module.exports = {
     return getCurrentDraft(bookingId)
       .then(form => ({
         id: form.id,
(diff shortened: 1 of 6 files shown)


integration-tests/db/db.js
```diff
@@ -22,13 +22,14 @@ module.exports = {
     return getCurrentDraft(bookingId)
       .then(form => ({
         id: form.id,
-        data: { incidentDate: form.incident_date, ...form.form_response.incident[formName] },
+        iincidentDate: form.incident_date,
```
jsrobertson: typo? `iincidentDate`

andrewrlee: thanks, fixed that

PR #122, opened 2019-08-23
DCS-85: Validation of fields in Relocation and Injuries page.

GurnankCheema:
http://localhost:3000/form/incident/relocationAndInjuries/-1

No validation previously. User will now be alerted to the fields that still require an input when they select the Save and Continue button.
Unit tests written to check functionality is correct.
Minor styling changes applied to margins around inputs.


server/views/formPages/incidentMacros.njk
```diff
@@ -184,53 +185,67 @@
...
+{% endif %}
+
+<div class="govuk-form-group {{govukFormGroupErrorOuter}}">
   <fieldset class="govuk-fieldset">
     <!-- primary -->
     <legend class="govuk-fieldset__legend govuk-fieldset__legend">
       {{question.primaryQuestion.text}}
     </legend>
+    <span class="govuk-error-message">
+      <span class="govuk-visually-hidden">Error:</span>
```
andrewrlee: isn't this always going to be present for screen readers?

server/views/formPages/incidentMacros.njk
```diff
@@ -184,53 +185,67 @@
...
+<!-- this one -->
 {% macro radiosWithNestedTextbox(question) %}
 
-<div class="govuk-form-group">
+{% if question.primaryQuestion.errorMessage.text %}
+{% set govukFormGroupErrorOuter  =  'govuk-form-group--error' %}
+{% set primaryErrorMessageText = question.primaryQuestion.errorMessage.text %}
+{% endif %}
+
+{% if not question.primaryQuestion.errorMessage and question.followUpQuestion.errorMessage.text %}
```
andrewrlee: not sure if the and is needed? if there is an error wont there always be text?

server/views/formPages/incidentMacros.njk
GurnankCheema: I didn't think it would be. I took it out of https://design-system.service.gov.uk/components/error-message/
and if you look at another example that doesn't have errors like https://design-system.service.gov.uk/components/fieldset/
, the govuk-visually-hidden is included there too. 
So my thoughts were that it will not be seen by screen readers unless it's active ?

PR #127, opened 2019-08-23
DCS-98 content changes for check your answers

andrewrlee:
* Fix incident type ‘was use of force planned’
 * involved staff per line
 * Witnesses per line
 * Was the prisoner compliant? Needs to be a question yes/no
 * Remove ‘-‘ where we use ‘and’ in questions
 * Number of officers should be an actual number rather than word
 * Staff taken to hospital - 'None' when none
 * Staff needed medical attention - 'None' when none


andrewrlee:
> Not sure what the **Remove ‘-‘ where we use ‘and’ in questions** means?

Oh, that's not very clear! - Andy told us to remove the dash where previously used the pattern`Yes - and was used`

PR #130, opened 2019-08-27
DCS-97 Moving url routes to the correct structure

andrewrlee:


server/views/pages/check-answers.html
```diff
@@ -187,7 +187,7 @@ <h1 class="govuk-heading-xl">Check your answers before sending the report</h1>
       pagesMacros.sectionHeading({
         data: data.evidence,
         title: 'Evidence',
-        url: '/form/incident/evidence/' + bookingId
+        url: '/report/' + bookingId + '/evidence'
```
petergphillips: would prefer string interpolation rather than just adding strings together

andrewrlee: I don't think nunjucks supports string interpolation?

PR #133, opened 2019-08-28
DCS-83 Adding incident details validation

andrewrlee:


server/config/incident.js
```diff
@@ -6,9 +6,21 @@ const removeEmptyValues = attrs => (inputs = []) => inputs.filter(hasAtLeastOneO
...
 
-const toBoolean = val => (val == null ? null : val === 'true')
+const toBoolean = val => {
+  switch (val) {
+    case 'true':
+      return true
+    case 'false':
+      return false
+    default:
+      return null
```
JamieDraperUK: Feels a bit wrong to be returning null here, I'm sure it makes sense given the right context.

andrewrlee: The form processing will remove the field if its null and not persist it in the DB. 
The changes here are to allow us to distinguish between true, false and ~FileNotFound~ not provided.

PR #135, opened 2019-08-29
DCS-78 a : User can now view additional comments entered previously

GurnankCheema:
1. New Knex migration to create the statement_amendments table in the DB
    2. Updated review.html to display the additional comments plus a new text box for future comments. 
    Changed Save buttons
    3 Updated unit tests for getStatement()
    4.Included new unit test for getAdditionalComments()

### DIFF
--- a/server/services/statementService.test.js
+++ b/server/services/statementService.test.js
@@ -7,6 +7,7 @@ const incidentClient = {
   getStatement: jest.fn(),
   saveStatement: jest.fn(),
   submitStatement: jest.fn(),
+  getAdditionalComments: jest.fn(),
 }
 
 let service
@@ -43,6 +44,8 @@ describe('getStatement', () => {
     expect(output).toEqual(incident)
 
     expect(incidentClient.getStatement).toBeCalledWith('BOB', 1, StatementStatus.PENDING)
+
+    expect(incidentClient.getAdditionalComments).toBeCalled()
   })
 
   test('retrieve details when statement is not present', async () => {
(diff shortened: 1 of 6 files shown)


server/views/pages/statement/your-statement.html
```diff
@@ -60,17 +60,58 @@ <h1 class="govuk-heading-xl mainHeading">Your use of force statement</h1>
...
+        text: 'Save additional comments',
+        name: 'saveAdditionalComments',
+        value: 'save-and-continue',
+        classes: 'govuk-button',
+        attributes: { 'data-qa': 'save-and-continue' }
+      })
+    }}
+    {{
+      govukButton({
+        text: 'Return to use of force statements',
```
andrewrlee: I think this can just be the same as it was before and just be a link back to the main statement page.

I don't think we will want to support an amendment lifecycle and support temporarily saving it before submitting? it would be confusing how to display that there is an incomplete amendment on the statement page.

GurnankCheema: updated as per comments, plus addition of     
<input type="hidden" name="_csrf" value="{{ csrfToken }}" />

server/services/statementService.test.js
```diff
@@ -43,6 +44,8 @@ describe('getStatement', () => {
     expect(output).toEqual(incident)
 
     expect(incidentClient.getStatement).toBeCalledWith('BOB', 1, StatementStatus.PENDING)
+
+    expect(incidentClient.getAdditionalComments).toBeCalled()
```
andrewrlee: should probably test the merging too: 

* rename `incident` -> `statement`
* add `incidentClient. getAdditionalComments.mockReturnValue([... one or more comments])`
* update `expect(output).toEqual(incident)` to `expect(output).toEqual({...statement, additionalComments: [...])`

GurnankCheema: done

PR #138, opened 2019-08-30
DCS-109 moving part of form json payload to use new section key

andrewrlee:
The json payload now refers to `useOfForceDetails` rather than the more vague `details`

There's no migration here, we're just going to wipe the data on dev as it's a bit out of date.


server/services/reportStatusChecker.test.js
```diff
@@ -2,7 +2,16 @@ const { SectionStatus, check } = require('./reportStatusChecker')
 
 describe('statusCheck', () => {
   const validReport = {
-    details: {
+    incidentDetails: {
+      witnesses: [{ name: 'BOB BARRY' }, { name: 'JAMES JOHN' }],
+      locationId: -25,
+      involvedStaff: [
+        { name: 'Itag User', email: 'itag_user@digital.justice.gov.uk', staffId: 1, username: 'ITAG_USER' },
```
petergphillips: out of interet why is `staffId` coming through here?  Are you using it at all - note that auth users won't have one...

andrewrlee: As discussed - I'll add a ticket to make this optional in the short term with the hope to move to a better required, immutable identifier in the future when one become available.

PR #140, opened 2019-08-30
DCS-78 - Users additional comments being persisted to db

GurnankCheema:
New functionality to take user's inputs in UI and persist to db.

User can navigate to the incident statement by clicking the View and Amend button on http://localhost:3000/

User can then input any additional details in the text box. Clicking the Save additional comments button persists new entry to db and redirects user back to the same screen. 

The additional comments are linked to the original statement via the statement.id value

Testing not yet included

### DIFF
--- a/integration-tests/integration/createIncident/save-additional-comment.spec.js
+++ b/integration-tests/integration/createIncident/save-additional-comment.spec.js
@@ -1,0 +1,68 @@
+const TasklistPage = require('../../pages/tasklistPage')
+const SubmittedPage = require('../../pages/submittedPage')
+const ViewStatementPage = require('../../pages/viewStatementPage')
+const IncidentsPage = require('../../pages/incidentsPage')
+
+context('Add comments to statement', () => {
+  const bookingId = 1001
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubLogin')
+    cy.task('stubOffenderDetails', bookingId)
+    cy.task('stubLocations', 'MDI')
+    cy.task('stubOffenders')
+    cy.task('stubLocation', '357591')
+    cy.task('stubUserDetailsRetrieval', 'Mr Zagato')
+    cy.task('stubUserDetailsRetrieval', 'Mrs Jones')
+    cy.task('stubUserDetailsRetrieval', 'Test User')
+  })
+
+  it('A user can select a specific statement, add to it and then return back to statements page', () => {
+    cy.login(bookingId)
+
+    const tasklistPage = TasklistPage.visit(bookingId)
+    tasklistPage.checkNoPartsComplete()
+
+    const newIncidentPage = tasklistPage.startNewForm()
+    newIncidentPage.fillForm()
+
+    const detailsPage = newIncidentPage.save()
+    detailsPage.fillForm()
+
+    const relocationAndInjuriesPage = detailsPage.save()
+    relocationAndInjuriesPage.fillForm()
+
+    const evidencePage = relocationAndInjuriesPage.save()
+    evidencePage.fillForm()
+
+    const checkAnswersPage = evidencePage.save()
+    checkAnswersPage.clickSubmit()
+
+    const submittedPage = SubmittedPage.verifyOnPage()
+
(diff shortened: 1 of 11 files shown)


integration-tests/integration/createIncident/save-additional-comment.spec.js
```diff
@@ -0,0 +1,67 @@
...
+    submitStatementPage.lastTrainingYear().type('2010')
+    submitStatementPage.jobStartYear().type('1999')
+    submitStatementPage.statement().type('This is my statement')
+
+    const confirmStatementPage = submitStatementPage.submit()
+    const statementSubmittedPage = confirmStatementPage.submit()
+    const incidentsPage = statementSubmittedPage.finish()
+    const { reportId } = incidentsPage.getCompleteRow(0)
+    reportId().then(id =>
+      cy.task('getStatementForUser', { reportId: id, status: StatementStatus.SUBMITTED }).then(() => {
```
andrewrlee: I don't think you need to call these methods - I think this is retrieving the statement data from the database and then just ignoring it.

Also don't need to drop down the raw cypress to click the view button (this assumes that there is only one page on the screen and so is prone to breakage). 

```javascript
  const { viewButton } = incidentsPage.getCompleteRow(0)
  viewButton().click()
```

integration-tests/integration/createIncident/save-additional-comment.spec.js
```diff
@@ -0,0 +1,67 @@
...
+        cy.get('.govuk-button').click()
+        cy.get('[id="additionalComment"]').type('Some new comment 1')
+        cy.get('[data-qa="save-and-continue"]').click()
+        cy.get('[data-qa="statement"]').contains('Some new comment 1')
```
andrewrlee: It would be better to use data-qa tags and the `row.index` to identify these uniquely - this test reads like it overrides the existing comment/statement.

Also these should live on the page object so the whole test is written at the same level of abstraction:
```
viewStatementPage.comment(1).should('not', 'exist')

viewStatementPage.newComment().type('aaaa')
viewStatementPage.comment(1).contains('aaaa')
viewStatementPage.save()

viewStatementPage.newComment().type('bbbb')
viewStatementPage.comment(2).contains('bbbb')
viewStatementPage.save()
viewStatementPage.newComment().should('be.empty')
```

integration-tests/integration/createIncident/save-additional-comment.spec.js
```diff
@@ -0,0 +1,67 @@
...
+        cy.get('[id="additionalComment"]').type('Lorum ipsum Lorum ipsum Lorum ipsum ')
+        cy.get('[data-qa="save-and-continue"]').click()
+        cy.get('[data-qa="statement"]').contains('Lorum ipsum Lorum ipsum Lorum ipsum')
+        cy.get('[id="additionalComment"]').should('be.empty')
+        cy.get('[data-qa="cancel"]').click()
+        cy.url().should('eq', 'http://localhost:3007/')
```
andrewrlee: There is a helper for this which is more explicit: `IncidentsPage.verifyOnPage()` - this will wait until the page title is as expected

integration-tests/integration/createIncident/save-additional-comment.spec.js
GurnankCheema: The reason I thought these methods were still required was, the test needs a form & statement to be created in order to then have access to the statement and add to it. 
In order for the test sequence to reach the statement, the form must pass all the preceding validations, hence the reason why the tests are included.
Although ... I may be incorrect in thinking this so I tried removing the preceding form completion and validations. Because there was nothing in the db, there was no reportId() to access

integration-tests/integration/createIncident/save-additional-comment.spec.js
GurnankCheema: used

integration-tests/integration/createIncident/save-additional-comment.spec.js
GurnankCheema: done

PR #144, opened 2019-08-30
DCS-111 strip out invalid fields 

andrewrlee:
in nested input and sanitise incoming strings by sanitising them


server/utils/fieldValidation.js
```diff
@@ -166,27 +159,26 @@ const fieldOptions = {
     })
   ),
 
-  requiredStaffNeedingMedicalAttention: () =>
```
andrewrlee: There are no changes in here apart from moving away from declaring these as functions where possible

PR #145, opened 2019-09-02
DCS-112 inlining remove buttons for add-another components

andrewrlee:
e.g: 

Previously:
<kbd>
Screenshot 2019-09-02 at 09 12 22
</kbd>

Now: 
<kbd>
Screenshot 2019-09-02 at 09 13 43
</kbd>

### DIFF
--- a/assets/js/add-another.js
+++ b/assets/js/add-another.js
@@ -3,8 +3,9 @@
  */
 var AddAnother = {}
 
-AddAnother = function(container) {
+AddAnother = function(container, removeContainerSelector) {
   this.container = $(container)
+  this.removeContainerSelector = removeContainerSelector
   this.container.on('click', '.add-another__remove-button', $.proxy(this, 'onRemoveButtonClick'))
   this.container.on('click', '.add-another__add-button', $.proxy(this, 'onAddButtonClick'))
   this.container.find('.add-another__add-button, add-another__remove-button').prop('type', 'button')
@@ -58,9 +59,11 @@ AddAnother.prototype.updateAttributes = function(index, item) {
 }
 
 AddAnother.prototype.createRemoveButton = function(item) {
-  item.append(
-    '<button type="button" class="govuk-button govuk-button--secondary add-another__remove-button">Remove</button>'
-  )
+  item
+    .find(this.removeContainerSelector)
+    .append(
+      '<button type="button" class="govuk-button govuk-button--secondary add-another__remove-button govuk-!-margin-left-3">Remove</button>'
+    )
 }
 
 AddAnother.prototype.resetItem = function(item) {
--- a/integration-tests/pages/evidencePage.js
+++ b/integration-tests/pages/evidencePage.js
@@ -25,7 +25,9 @@ export default () =>
       cy.get('[name="bodyWornCameraNumbers[1][cameraNum]"]').type('456')
       cy.get('[data-qa-add-another-camera = true]').click()
       cy.get('[name="bodyWornCameraNumbers[2][cameraNum]"]').type('789')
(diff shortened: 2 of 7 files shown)


assets/js/add-another.js
```diff
@@ -3,8 +3,9 @@
  */
 var AddAnother = {}
 
-AddAnother = function(container) {
+AddAnother = function(container, removeContainerSelector) {
   this.container = $(container)
+  this.removeContainerSelector = removeContainerSelector
   this.container.on('click', '.add-another__remove-button', $.proxy(this, 'onRemoveButtonClick'))
   this.container.on('click', '.add-another__add-button', $.proxy(this, 'onAddButtonClick'))
   this.container.find('.add-another__add-button, add-another__remove-button').prop('type', 'button')
```
JamieDraperUK: Why is this required?

andrewrlee: Previously, the add another component would clone the element and then append a remove button to it. This provides more control so we can choose where exactly the remove button is placed

assets/js/add-another.js
```diff
@@ -58,9 +59,11 @@ AddAnother.prototype.updateAttributes = function(index, item) {
 }
 
 AddAnother.prototype.createRemoveButton = function(item) {
-  item.append(
-    '<button type="button" class="govuk-button govuk-button--secondary add-another__remove-button">Remove</button>'
-  )
+  item
+    .find(this.removeContainerSelector)
+    .append(
+      '<button type="button" class="govuk-button govuk-button--secondary add-another__remove-button govuk-!-margin-left-3">Remove</button>'
```
JamieDraperUK: Not a massive fan of sticking HTML in strings like this as they tend to get missed by lint like tooling etc. Maybe, tackle it if we end up doing more of it?

integration-tests/pages/evidencePage.js
```diff
@@ -25,7 +25,9 @@ export default () =>
       cy.get('[name="bodyWornCameraNumbers[1][cameraNum]"]').type('456')
       cy.get('[data-qa-add-another-camera = true]').click()
       cy.get('[name="bodyWornCameraNumbers[2][cameraNum]"]').type('789')
-      cy.get('.add-another-camera > :nth-child(2) > .govuk-button').click()
+      cy.get('.add-another-camera .add-another__remove-button')
```
JamieDraperUK: Feels a bit inconsistent here, above we've not tied ourselves to the document structure whereas for the camera buttons we have, could make for a brittle test?

server/views/formPages/incident/evidence.html
```diff
@@ -335,7 +293,7 @@ <h1 class="govuk-heading-xl mainHeading">Evidence</h1>
   <script src="/assets/js/jquery.min.js"></script>
   <script src="/assets/add-another.js"></script>
   <script>
```
JamieDraperUK: Do we need to do a document ready here? or is the order of the scripts enough?

assets/js/add-another.js
andrewrlee: yes, we've inherited this from the hmcts add another.. will consider this if we need to do more of it.

integration-tests/pages/evidencePage.js
andrewrlee: I agree, I think this is a major improvement from before which relied on a specific relationship between 3 elements - this also is more descriptive as at least we know which button is being clicked. I've refactored the other page objects to be a bit more explicit in structure but haven't got to this one yet

server/views/formPages/incident/evidence.html
andrewrlee: Not sure? we haven't had any issues with this so far

PR #154, opened 2019-09-06
DCS-118: User is now able to view the current statements before movin…

GurnankCheema:
…g to an Add Comments page

http://localhost:3000/reportId/add-comment-to-statement

The previous journey was for user to view and amend the Statement on the same page.
The button to access the Statement page was on the Incidents page
The journey has been changed so that the the Incidents page only has a view button. This takes user to the Statement page which displays the initial statement plus any additional comments.
From here the user can either click Continue to go back to the Incidents page or click Add comment to go to the new page.
The new page displays the same information as the Statement page but also has a text box to accept another comment.
The user then has the option to either Save a new comment (and go back to the Incidents page) or Cancel and return to the previous page.

### DIFF
--- a/server/routes/incidents.js
+++ b/server/routes/incidents.js
@@ -154,7 +154,7 @@ module.exports = function CreateReportRoutes({ statementService, offenderService
       return res.redirect(`/`)
     },
 
-    addCommentToStatement: async (req, res) => {
+    viewAddCommentToStatement: async (req, res) => {
       const { reportId } = req.params
       const statement = await statementService.getStatement(req.user.username, reportId, StatementStatus.SUBMITTED)
       const offenderDetail = await offenderService.getOffenderDetails(res.locals.user.token, statement.bookingId)
--- a/server/routes/incidents.js
+++ b/server/routes/incidents.js
@@ -151,7 +151,27 @@ module.exports = function CreateReportRoutes({ statementService, offenderService
       if (req.body.additionalComment && req.body.additionalComment.trim().length) {
         await statementService.saveAdditionalComment(statement.id, req.body.additionalComment)
       }
-      return res.redirect(`/${reportId}/your-statement`)
+      return res.redirect(`/`)
+    },
+
+    addCommentToStatement: async (req, res) => {
+      const { reportId } = req.params
+      const statement = await statementService.getStatement(req.user.username, reportId, StatementStatus.SUBMITTED)
+      const offenderDetail = await offenderService.getOffenderDetails(res.locals.user.token, statement.bookingId)
+      const { displayName, offenderNo } = offenderDetail
+
+      if (req.body.additionalComment && req.body.additionalComment.trim().length) {
+        await statementService.saveAdditionalComment(statement.id, req.body.additionalComment)
+      }
+      return res.render('pages/statement/add-comment-to-statement', {
+        data: {
+          reportId,
+          displayName,
+          offenderNo,
+          ...statement,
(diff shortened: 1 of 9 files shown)


server/routes/incidents.js
```diff
@@ -151,7 +151,27 @@ module.exports = function CreateReportRoutes({ statementService, offenderService
       if (req.body.additionalComment && req.body.additionalComment.trim().length) {
         await statementService.saveAdditionalComment(statement.id, req.body.additionalComment)
       }
-      return res.redirect(`/${reportId}/your-statement`)
+      return res.redirect(`/`)
+    },
+
+    addCommentToStatement: async (req, res) => {
```
andrewrlee: can we call this `viewAddCommentToStatement` so it matches the same convention as the other routes?

server/routes/incidents.test.js
```diff
@@ -110,13 +110,13 @@ describe('GET /:reportId/your-statement', () => {
       }))
 })
 
-describe('POST /:reportId/your-statement', () => {
-  it('should save ammendment', () =>
+describe('POST /:reportId/add-comment-to-statement', () => {
```
andrewrlee: think you need a test for the  `GET` add-comment-to-statement

server/views/pages/statement/your-statement.html
```diff
@@ -55,69 +54,49 @@ <h1 class="govuk-heading-xl mainHeading">Your use of force statement</h1>
   </div>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      <p class="govuk-!-margin-bottom-1 govuk-!-margin-top-0 govuk-label--xs statement" data-qa="statement">{{ data.statement }}  </p>
+      <p class="govuk-!-margin-bottom-1 govuk-!-margin-top-0 govuk-label--xs statement" data-qa="statement">
+        {{ data.statement }} </p>
```
andrewrlee: think this is going to add extra padding to the start of the statement as it has `white-space: pre-wrap` - needs to be on one line

server/views/pages/statement/add-comment-to-statement.html
```diff
@@ -0,0 +1,122 @@
...
+    <div class="govuk-grid-column-one-half ">
+      <p>
+        <span class="govuk-label--s">Submitted:</span> {{ data.submittedDate | formatDate('D MMMM YYYY - HH:mm') }}
+      </p>
+    </div>
+  </div>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <p class="govuk-!-margin-bottom-1 govuk-!-margin-top-0 govuk-label--xs statement" data-qa="statement">
+        {{ data.statement }} </p>
```
andrewrlee: hink this is going to add extra padding to the start of the statement as it has white-space: pre-wrap - needs to be on one line

server/routes/incidents.js
GurnankCheema: yes

server/views/pages/statement/your-statement.html
GurnankCheema: Yes, that happened when I used the linter. 
Amended

server/views/pages/statement/add-comment-to-statement.html
GurnankCheema: as above.

server/routes/incidents.test.js
GurnankCheema: done

PR #155, opened 2019-09-06
DCS-93 Processing reminders

andrewrlee:
The logic to work out which type of reminder to send based on whether the recipients
was the reporter and whether they are overdue or not.

I've also updated the email templates and added an environmental variable for the link
which appears in the emails.

After being processed, items should become ineligible for subsequent polls by either
having their next reminder date push to the next day, or if they statement is overdue
then the date will be blanked out to show that no further reminders will occur.

I've temporarily reduced the amount of items that are processed in one go, as don't potentially
want 100s of emails being sent whilst testing.


server/services/notificationService.js
```diff
@@ -13,52 +20,94 @@ const createNotificationService = emailClient => {
...
           REPORTER_NAME: reporterName,
           INCIDENT_DATE: asDate(incidentDate),
           INCIDENT_TIME: asTime(incidentDate),
           DEADLINE_DATE: asDeadline(incidentDate),
+          LINK: emailUrl,
+        },
+        reference: null,
+      })
+      .then(({ body }) => logger.info(`Send statement reminder, successful for reporter: '${reporterName}'`, body))
+      .catch(({ message }) => logger.error(`Send statement reminder, failed for reporter: '${reporterName}'`, message))
```
petergphillips: is it right that we don't throw exception on failure to send email?
Is the logger monitored for failures or would this better to be a custom event in analytics?
How about throwing exception and then setting the next reminder to be the following day instead?

andrewrlee: We do have a story to add events to analytics - as we will want to track both successes and failures of this. 

I'll have a think about the error handling - I mostly didn't want to block the queue if an error sending emails happens - though as you say, catching the error higher up and giving another chance to send the 'overdue' email might be a good move

PR #164, opened 2019-09-10
DCS-76 Adding 'Your report' page

andrewrlee:
Allowing users to view their submitted reports.
This reuses a lot of the logic that the check your answer page uses. 

Also removed references to 'My ...' to 'Your ...'


server/routes/incidents.js
```diff
@@ -17,26 +19,47 @@ module.exports = function CreateReportRoutes({ reportService, offenderService })
...
         completedReports,
-        selectedTab: 'my-reports',
+        selectedTab: 'your-reports',
       })
     },
+
+    viewYourReport: async (req, res) => {
+      const { reportId } = req.params
+
+      const { id, form = {}, incidentDate, bookingId } = await reportService.getReport(req.user.username, reportId)
```
JamieDraperUK: Might be worth checking user input here and return a bad request etc.

JamieDraperUK: Is there a general catch-all error in place that returns a standard error template?

andrewrlee: 👍 We do need to check this. We don't have an error page yet - we have a story for it.

PR #169, opened 2019-09-13
DCS-129 Adding middleware to detect users who have the reviewer role

andrewrlee:
Now JWT Token has to be an actual valid value - so generating a proper one


server/middleware/authorisationMiddleware.js
```diff
@@ -0,0 +1,11 @@
+const jwtDecode = require('jwt-decode')
+
+module.exports = (req, res, next) => {
+  if (res.locals && res.locals.user && res.locals.user.token) {
```
Robc24: Do you need to check for `**res.locals.user**` here if you're checking for `**res.locals.user**.token`?

andrewrlee: This is based on something I nabbed from the cat tool 😄 

I would have thought so though, if there's no user present on locals - then wouldn't it blow up with `no field token on undefined`?

server/middleware/authorisationMiddleware.js
```diff
@@ -0,0 +1,11 @@
...
+    const roles = jwtDecode(res.locals.user.token).authorities
+    res.locals.user = { isReviewer: roles.includes('ROLE_USE_OF_FORCE_REVIEWER'), ...res.locals.user }
```
bell-pa: could roles be null?

andrewrlee: I'm now handling this and have improved tests - generating a token rather than hardcoding them in there

PR #173, opened 2019-09-16
Dcs 123 'write your statement' accepts 123456 as a valid year but blows up (UoF)

bell-pa:
Also noticed that the constraint on month isn't linked to year. For example, when the date is in September 2019 then an input of 'Month: September, Year: 2019' should be valid, but 'Month: October, Year: 2019' should not.

Fixed the 123456 problem by reducing the range of the 'sanitiser' to that of PostgreSQL's smallInt (-32768 .. 32767)

Fixed the month - year cross-field validation problem by introducing a new requiredMonthIndexNotInFuture function in fieldValidation.js and using this in statement.js. (See new tests in fieldValidation.test.js)


server/config/statementValidation.test.js
```diff
@@ -106,6 +111,10 @@ describe('lastTrainingYear', () => {
         href: '#lastTrainingYear',
         text: 'Enter the year you last attended refresher training',
       },
+      {
```
andrewrlee: isn't a month present in both these last two test cases?

bell-pa: Yes it is, but it is month 11, while this month is 8.  When the requiredMonthIndexNotInFuture constraint determines that the year is not in the past it insists that month must be this month or earlier. The year strings 'asasas' and  `${moment().year() + 1}` are treated as being the past hence the month error message.  Perhaps there is a way of supressing month related errors when there are errors for the year.  Alternatively the month and year could be treated as one field for validation purposes...

andrewrlee: as discussed, we'll raise a bug for this.. it's a relatively minor issue

PR #174, opened 2019-09-16
DCS-129 Storing agency on the report level

andrewrlee:
To allow us to filter by agencies that reviewers are allowed to see

Also adding an updated_date to reports and setting when updating reports. 

Putting in place an overdue date column on statements which will allow us to easily query if statements are overdue or not

### DIFF
--- a/server/data/incidentClient.test.js
+++ b/server/data/incidentClient.test.js
@@ -83,6 +83,7 @@ test('createDraftReport', async () => {
   const id = await incidentClient.createDraftReport({
     userId: 'user1',
     bookingId: 'booking-1',
+    agencyId: 'LEI',
     reporterName: 'Bob Smith',
     offenderNo: 'AA11ABC',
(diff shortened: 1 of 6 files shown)


server/data/incidentClient.test.js
```diff
@@ -119,11 +125,14 @@ test('submitReport', () => {
...
-    where user_id = $3
-    and booking_id = $4
-    and status = $5
-    and r.sequence_no = (select max(r2.sequence_no) from report r2 where r2.booking_id = r.booking_id and user_id = r.user_id)`,
+    text: `update report r
+            set status = $1
+            ,   submitted_date = $2
+            ,   updated_date = now()
+          where user_id = $3
+          and booking_id = $4
```
lbennett-moj: v minor - consistency with the table alias?

andrewrlee: Thanks, I'll make consistent in a future PR.

PR #179, opened 2019-09-16
DCS-126: Top half of Write UOF statement reformatted

GurnankCheema:
The content is still the same as before. The only differences are:
1, the Year field for joining Prison Service has been dropped to a new row below the month and Year of Refresher training
2, have removed the hardcoded h1 title and pushed the page title from the back-end into both the current page and the layout.html. This will cause the browser tab to also displays the page title.

### DIFF
--- a/integration-tests/integration/statements/submit-statement.spec.js
+++ b/integration-tests/integration/statements/submit-statement.spec.js
@@ -42,7 +42,7 @@ context('Submit statement', () => {
     }
 
     const submitStatementPage = SubmitStatementPage.verifyOnPage()
-    submitStatementPage.offenderName().contains('Norman Smith (A1234AC)')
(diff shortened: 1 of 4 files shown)


integration-tests/integration/statements/submit-statement.spec.js
```diff
@@ -42,7 +42,7 @@ context('Submit statement', () => {
     }
 
     const submitStatementPage = SubmitStatementPage.verifyOnPage()
-    submitStatementPage.offenderName().contains('Norman Smith (A1234AC)')
+    submitStatementPage.offenderName().contains('Smith, Norman')
```
andrewrlee: Andy said that it should be first name last name, any place where it isn't a list

PR #181, opened 2019-09-17
DCS-129 Filtering all incidents by current users caseload

andrewrlee:


server/routes/incidents.js
```diff
@@ -33,7 +33,7 @@ module.exports = function CreateReportRoutes({ reportService, involvedStaffServi
         return res.redirect('/your-statements')
       }
 
-      const { awaiting, completed } = await reportService.getReportsForReviewer()
+      const { awaiting, completed } = await reportService.getReportsForReviewer(res.locals.user.activeCaseLoadId)
```
bell-pa: Is there always a res.locals.user.activeCaseLoadId ?

andrewrlee: The use-of-force reviewer should always be viewing the reports in the context of a prison.

PR #193, opened 2019-09-23
DCS-127 Fixing statement buttons

andrewrlee:
Showing 'start statement' or 'continue statement' depending on whether revisiting an edited statement or not

Also fixing display of buttons and labels - so consistent with other buttons and labels within the same page.


migrations/20190923000000_form_db.js
```diff
@@ -0,0 +1,16 @@
+exports.up = knex =>
+  Promise.all([
+    knex.schema.table('statement', table => {
```
bell-pa: Looks like Promise.all isn't needed here...

PR #213, opened 2019-10-04
Adding new-nomis docker compose file

andrewrlee:
Also:
* Updating README.md with info about accessing the application.
* Fixing audit issues


docker-compose.newnomis.yml
```diff
@@ -0,0 +1,22 @@
...
+      - oauth-server
+      - elite2-api
+    container_name: new-nomis-ui
+    ports:
+      - '3001:3001'
+    healthcheck:
+      test: ['CMD', 'curl', '-f', 'http://localhost:3001/health']
+    environment:
+      - PORT=3001
+      - NODE_ENV=dev
```
petergphillips: convention dictated that this should either be `development` or `production`

docker-compose.newnomis.yml
```diff
@@ -0,0 +1,22 @@
...
+  new-nomis-ui:
+    image: mojdigitalstudio/new-nomis-ui:latest
+    networks:
+      - hmpps
+    depends_on:
+      - oauth-server
+      - elite2-api
+    container_name: new-nomis-ui
+    ports:
+      - '3001:3001'
```
petergphillips: I think it would be much simpler if this was kept as 3000 and use of force changed, otherwise running nomis, PSH locally just gets annoying.  We have tried to keep nomis, PSH and omic as 3000, 3002, 3001 and others use unique ports above that...

andrewrlee: I'm guessing we don't have a list of unique ports which make choosing a unique one challenging... also sentence-planning, pathfinder and cat-tool don't seem to follow this convention and all run on port 3000 by default

docker-compose.newnomis.yml
andrewrlee: fixed this - think this was copied over from pathfinder.

PR #218, opened 2019-10-07
DCS-160: fixed blue section in page header

GurnankCheema:
The blue section was previously slightly taller than the surrounding black bar so have reduced its height. Have also chaned the layout when screen is Samsung S5 sized. The only 2 screen sizes needed to be considered for are Samsung S5 and desktop
Note have changed the blue colour to shade #015ea5 to be consistent with DPS. Could not find a sass code equivalent to #015ea5 in the latest GDS toolkit hence have used the hex code

### DIFF
--- a/assets/sass/local.sass
+++ b/assets/sass/local.sass
@@ -12,6 +12,7 @@ div.govuk-width-container
 div.govuk-header__container
   border-bottom: 0
   padding-top: 0
+  padding-right: 0
 
 div.govuk-header__logo
   padding-top: 10px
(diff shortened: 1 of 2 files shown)


assets/sass/local.sass
```diff
@@ -27,17 +28,25 @@ header.govuk-header
...
+    border-right: none
+  @include govuk-media-query($from: 360px)
+    border-right: solid white 1px
 
 .user-block
   color: white
   float: right
   font-size: 19px
-  background-color: govuk-colour("blue")
+  background-color: #015ea5
```
bell-pa: Bespoke colours like this are probably due to Eddie ignoring the Gov UK palette.  I'd suggest introducing a SASS variable for this colour.  That would make it easy to change its value later, and also makes it clear that all the places it is used are sharing the same value.  There are two occurrences in this file already.

GurnankCheema: As suggested, have created a new file called palette.sass and defined the variable $govuk-blue for use in the local.sass
Have also updated the media queries from using actual px to $from: mobile and $from: desktop 
Lastly, have changed the padding in the ul and li so that links can all fit on one line when screen is in mobile size

PR #225, opened 2019-10-15
Alee/dcs 155 remove cls hooked

andrewrlee:


job/reminders/reminderPoller.js
```diff
@@ -15,7 +15,7 @@ module.exports = (db, incidentClient, sendReminder, eventPublisher) => {
     let count = 0
 
     while (result && count < 50) {
-      result = await db.inTransaction(processReminder)
+      result = await db.inTransaction(client => processReminder(client))
       count += result ? 1 : 0
```
bell-pa: Not sure this change is necessary.  db.inTransaction(processReminder) should be sufficient...

andrewrlee: fixed this.. and also added query to transaction

PR #230, opened 2019-10-20
DCS-155 Adding redis as a session store

andrewrlee:
Moving away from using the cookie to store the session and instead storing the session in redis.

This will allow larger amounts of data to be stored in flash state 

At the moment - if the user attempts to submit a partially valid form with lots of input, then:
 * The app attempts to set a session cookie that is too large for the browser and a new signature cookie for it
 * The browser refuses it but sets the new .sig cookie
 * When the browser returns the old session cookie and new .sig cookie it fails validation and assumes the user is unauth'd
 * This triggers redirects to `app/logout` -> `oath/token` -> `app/login` -> `app/page`  which creates an entirely new session cookie with no flash state
 * This means that the user is presented the same page but with no content/error messages.


andrewrlee:
Adding do not merge - as this could potentially go bad on the deploy to dev and I'm away next week so won't be able to support.
(Still need to roll out the necessary infrastructure changes across to pre-prod and prod)

docker-compose.test.yml
```diff
@@ -14,6 +14,15 @@ services:
     - POSTGRES_USER=use-of-force
     - POSTGRES_DB=use-of-force-int
 
+  use-of-force-redis:
+    image: 'bitnami/redis:5.0'
+    networks:
+      - hmpps
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
```
lbennett-moj: better var name than mine!

PR #239, opened 2019-11-27
P.O.C of sharing whitelists

andrewrlee:
Suspect this is unnecessarily complex - the original idea was to share whitelists across applications but quickly realised they were very diverse so reusing chunks of whitelists might be better.

Other much simpler solutions would be:
 * No whitelists
 * 1 single whitelist we use for everything
 * 3 whitelists, one per env type

We would have a global shared `values-ips.yml` which would be pulled in the build at deploy. This defines a number of named list of ips to whitelist:
```yml
whitelist_groups:

  internal:
    cloudplatform-live1-1: 35.178.209.113/32
    cloudplatform-live1-2: 3.8.51.207/32
    cloudplatform-live1-3: 35.177.252.54/32
    ...

 prisons:
    dxc-mitcheldean: 195.92.38.16/28
    ...
```

Then in each env specific`values.yml` file in each project, a list of whitelists can be specified:
```yml
whitelists:
  - internal
  - prisons 
```
  
I also wrote a [JS script](https://github.com/andrewrlee/k8-ip-whitelist-check) to allow inspecting the changes to the ip whitelists before applying.

Though in hindsight, I think there are more [generic helm extensions](https://github.com/databus23/helm-diff) that allow you to do similar things.

### DIFF
--- a/helm_deploy/values-ips.yml
+++ b/helm_deploy/values-ips.yml
@@ -1,0 +1,49 @@
+whitelist_groups:
+
+  internal:
+    health-kick: 35.177.252.195/32
+    mojvpn: 81.134.202.29/32
+    office: 217.33.148.210/32
+    petty_france_wifi: 213.121.161.112/28
+  
+  cloudplatform:
+    cloudplatform-live1-1: 35.178.209.113/32
+    cloudplatform-live1-2: 3.8.51.207/32
+    cloudplatform-live1-3: 35.177.252.54/32
+
+  ark:
+    ark-dom1-corsham-psn: 51.247.4.0/24
+    ark-dom1-corsham: 194.33.196.0/24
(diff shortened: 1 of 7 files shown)


helm_deploy/values-ips.yml
```diff
@@ -0,0 +1,49 @@
...
+    sodexo1: 80.86.46.16/32
+    sodexo2: 80.86.46.17/32
+    sodexo3: 80.86.46.18/32
+
+  pentesting:
+    pentester-bsi-1: 145.239.254.21/32
+    pentester-bsi-2: 54.37.241.156/30
+ 
+  other: 
+    durham-tees-valley: 51.179.193.241/32
```
andrewrlee: this file is arbitrarily structured. We've removed the pentesting whitelisting from cat tool since, so it's a bit out of date

PR #240, opened 2019-12-10
DCS-249 Splitting out user and system elite clients. 

andrewrlee:
This is to allow reviewers and reporters to view pull offender details of users who they have created reports for. Currently if the prisoner leaves the prison where the report was raised - they stop having the ability to access that prisoners details. This now uses client credentials when retrieving the offender details to by pass those checks

This will require adding the system_user role and the ability to use the client credentials grant type to the existing use-of-force credentials. 

Also had to swap the DB ports for local development and integration tests around - as the circle-ci build image seems to have randomly stopped respecting the port override config on our integration tests.


package.json
```diff
@@ -6,16 +6,16 @@
   "license": "MIT",
   "scripts": {
     "start": "node server.js",
-    "start:dev": "npm run build && DEBUG=gov-starter-server* nodemon server.js | bunyan -o short",
-    "start-feature": "NOTIFY_ENABLED=false PORT=3007 NOMIS_AUTH_URL=http://localhost:9091/auth ELITE2API_ENDPOINT_URL=http://localhost:9091 DB_NAME=use-of-force-int DB_PORT=5433 NODE_ENV=development node $NODE_DEBUG_OPTION server.js | bunyan -o short",
-    "start-feature:dev": "NOTIFY_ENABLED=false PORT=3007 NOMIS_AUTH_URL=http://localhost:9091/auth ELITE2API_ENDPOINT_URL=http://localhost:9091 DB_NAME=use-of-force-int DB_PORT=5433 NODE_ENV=development nodemon $NODE_DEBUG_OPTION server.js | bunyan -o short",
+    "start:dev": "npm run build && DEBUG=gov-starter-server* DB_PORT=5433 nodemon server.js | bunyan -o short",
```
brightonsbox: Just checking this should be 5433 and not 5432?

andrewrlee: yes - we have two postgres DBs - one for test and one for dev so they don't interfere.
test ran on 5433 previously but we've had to swap it to 5432 and so dev has taken over 5433.
It's annoying for various reasons but easiest fix for now

PR #252, opened 2020-02-03
DCS-336 Initial work to allow soft delete of reports

andrewrlee:
Added a new `deleted` columns and views for each table. 
Updated unique report index per booking/user to ignore reports which are deleted. 
Also using views when selecting or updating single table and using original tables when joining.


JamieDraperUK:
LGTM, however I have a strong urge to put all sql queries into named variables. It's just sql everywhere.

PR #253, opened 2020-02-03
DCS-336 Adding frontend to allow coordinators to soft delete reports

andrewrlee:
A delete button appears on the view-report screen if the current user is a coordinator.
This goes to a confirmation page and clicking the button there will trigger soft delete which redirects to the home page.

Need to get content folk to look at it and suggest better text/interaction design.

Also fixing new system token bug - was trying to read an unknown property called token from the token (rather than access_token)

### DIFF
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -3,13 +3,7 @@ const path = require('path')
 const httpError = require('http-errors')
 const asyncMiddleware = require('../middleware/asyncMiddleware')
 
-module.exports = function Index({
-  authenticationMiddleware,
-  offenderService,
-  reportingService,
-  involvedStaffService,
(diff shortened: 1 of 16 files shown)


server/routes/api.js
```diff
@@ -79,7 +55,7 @@ module.exports = function Index({
       const agencyId = res.locals.user.activeCaseLoadId
 
       const results = await reportingService.getMostOftenInvolvedPrisoners(
-        systemToken(res.locals.user.username),
+        await systemToken(res.locals.user.username),
```
andymarke: A bit confused how this worked before without the `await`

andrewrlee: correct!, it didn't!

PR #255, opened 2020-02-17
DCS-374 Allow creating of reports with unverified emails

andrewrlee:
Rather than preventing unverified users from completing reports we will mark them as unverfiied in the json blob.

We now add an additional interrupt page after involved staff who are unverified are added to a report:

Screenshot 2020-02-24 at 08 57 49

We also provide information to use of force reviewers and coordinators if a member of staff hasn't verified their email:

Screenshot 2020-02-22 at 10 32 55

This also changes the reminder chase job - if a reminder would be processed but the user is unverified, we'll do an additional call to auth service to see if the user has verified their account since the report was created, this will persist their email and also actually send them the reminder email.


package.json
```diff
@@ -75,10 +75,14 @@
     }
   },
   "lint-staged": {
-    "*.{js,json,css},!package.json,!package-lock.json": [
+    "*.{js,css}": [
       "prettier --write",
       "eslint --fix",
       "git add"
+    ],
+    "*.{json}": [
```
andrewrlee: the glob was previously broken so not picking up files

integration-tests/pages/createReport/userNotVerifiedPage.js
```diff
@@ -0,0 +1,10 @@
+const page = require('../page')
+
+const userNotVerifiedPage = username =>
+  page(`${username} name has not verified their email address`, {
```
bell-pa: Is this right? `${username} name has not verified their email address`
Should it be `${username} has not verified their email address` perhaps?

PR #272, opened 2020-03-12
Adding ability for coordinators to add missing staff to reports

andrewrlee:
If a new staff member is added to a completed report it will mark the report as incomplete.

Staff are added via the report details screen:
<kbd>Screenshot 2020-03-12 at 16 02 31</kbd>

One a user has been added they will be presented the "view report" screen.

There are 3 different error cases that are handled and present an interrupt screen with an appropriate message:
1.) The user does not exist: will lead the user back to 'add another user' screen to try again
2.) The user is already on the report: will lead the user back to "view report"
3.) The user is unverified: this adds the user to the report and will redirect to 'view report'

Also needed to fix a unique index on statements which ensured a user can appear on a report once. It now will allow users to have more than one statement on a report - as long as only a single report is not deleted.

### DIFF
--- a/integration-tests/integration/coordinator/remove-involved-staff.spec.js
+++ b/integration-tests/integration/coordinator/remove-involved-staff.spec.js
@@ -39,7 +39,7 @@ context('A use of force coordinator can remove involved staff', () => {
       ],
     })
 
-  it('A reviewer can remove staff on otherwise complete report and it will complete the report', () => {
+  it('A coordinator can remove staff on otherwise complete report and it will complete the report', () => {
     cy.task('stubCoordinatorLogin')
     cy.login(bookingId)
 
@@ -57,8 +57,10 @@ context('A use of force coordinator can remove involved staff', () => {
     }
 
     let viewStatementsPage = ViewStatementsPage.verifyOnPage()
-    viewStatementsPage.getReportId().then(reportId => cy.task('submitStatement', { userId: 'TEST_USER', reportId }))
-    cy.reload()
+    viewStatementsPage
+      .getReportId()
+      .then(reportId => cy.task('submitStatement', { userId: 'TEST_USER', reportId }))
+      .then(() => cy.reload())
 
     viewStatementsPage
       .statements()
--- a/server/routes/coordinator.js
+++ b/server/routes/coordinator.js
@@ -1,27 +1,61 @@
-const httpError = require('http-errors')
+const { AddStaffResult } = require('../services/involvedStaffService')
+const { firstItem } = require('../utils/utils')
 
 module.exports = function Index({ reportService, involvedStaffService, reviewService, offenderService, systemToken }) {
-  /**
-   * TODO: integrate this into the app properly once we have designs and time.
-   * This is currently a GET to get around CSRF issues and to allow access via a browser using current authenticated session :-/
-   */
   return {
-    addInvolvedStaff: async (req, res) => {
-      if (!res.locals.user.isCoordinator) {
-        throw httpError(401, 'Not authorised to access this resource')
+    viewAddInvolvedStaff: async (req, res) => {
+      const { reportId } = req.params
+
+      const errors = req.flash('errors')
+      const data = { incidentId: reportId }
+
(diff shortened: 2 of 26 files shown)


integration-tests/integration/coordinator/remove-involved-staff.spec.js
```diff
@@ -108,33 +110,33 @@ context('A use of force coordinator can remove involved staff', () => {
       )
   })
```
bell-pa: Delete commented test?

andrewrlee: Re-added that test back in 

An issue with cypress is that you can't just run a single test in isolation so I often end up commenting out others in the same suite whilst developing a new test

server/routes/coordinator.js
```diff
@@ -1,27 +1,61 @@
...
-    confirmDeleteReport: async (req, res) => {
-      if (!res.locals.user.isCoordinator) {
-        throw httpError(401, 'Not authorised to access this resource')
+    viewAddInvolvedStaffResult: async (req, res) => {
+      const { reportId, result } = req.params
+      const username = firstItem(req.flash('username'))
+
+      const knownResult = Object.values(AddStaffResult).some(knownValue => knownValue === result)
+
+      if (!username || !knownResult || result === AddStaffResult.SUCCESS) {
```
andrewrlee: As we store the username in flash, we won't have the username after a refresh of this page.
In that case, think it makes sense to just redirect to report view so they can see current state of the users.

server/routes/index.js
```diff
@@ -66,70 +67,86 @@ module.exports = function Index({
     next()
   })
 
-  const reportPath = formName => `/report/:bookingId/${formName}`
+  // All users
+  {
```
bell-pa: Consider extracting to functions with meaningful names (to eliminate comments)

server/services/involvedStaffService.js
```diff
@@ -2,159 +2,180 @@ const moment = require('moment')
...
-    await incidentClient.updateDraftReport(id, null, updatedFormObject)
-  }
-
-  const getInvolvedStaff = reportId => incidentClient.getInvolvedStaff(reportId)
+const AddStaffResult = {
+  SUCCESS: 'success',
+  SUCCESS_UNVERIFIED: 'unverified',
+  MISSING: 'missing',
+  ALREADY_EXISTS: 'already-exists',
+}
```
bell-pa: add `.freeze()` ?   Not really needed, but emphasises that these are immutable values.  I should do this myself!

server/routes/coordinator.js
```diff
@@ -1,27 +1,61 @@
...
-  /**
-   * TODO: integrate this into the app properly once we have designs and time.
-   * This is currently a GET to get around CSRF issues and to allow access via a browser using current authenticated session :-/
-   */
   return {
-    addInvolvedStaff: async (req, res) => {
-      if (!res.locals.user.isCoordinator) {
-        throw httpError(401, 'Not authorised to access this resource')
+    viewAddInvolvedStaff: async (req, res) => {
+      const { reportId } = req.params
```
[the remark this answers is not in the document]
andrewrlee: We actually have a jira ticket for validating request params generally.

server/views/pages/reportDetailMacro.njk
```diff
@@ -44,7 +44,7 @@
   </div>
 {% endmacro %}
 
-{% macro detail(data, bookingId, user) %}
+{% macro detail(data, bookingId, incidentId, user) %}
     {{
```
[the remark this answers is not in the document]
andrewrlee: theres a bit of gnarliness here around here and could do with a bit of tidying up

server/routes/index.js
andrewrlee: I was torn.. it's top level of the module so it looked a bit weird..

I'll address this in a follow up

server/services/involvedStaffService.js
andrewrlee: I have done that before.. I'll do in a follow up

PR #275, opened 2020-03-13
DCS-236 'Was the prisoner compliant' question enhanced

GurnankCheema:
http://localhost:3000/report/-19/relocation-and-injuries

The question has been enhanced so that if the user selects No, they are presented with 3 radios buttons each of which relate a different question.

If they had selected 'Yes' then there would be no supplementary radios.

The 'check your answers'  and the 'your-report' also
display the choices made to the supplementary question.


server/routes/model/reportSummary.js
```diff
@@ -41,7 +41,8 @@ const createUseOfForceDetails = (details = {}) => {
 const createRelocation = (relocationAndInjuries = {}) => {
   return {
     prisonerRelocation: toLabel(RelocationLocation, relocationAndInjuries.prisonerRelocation),
-    prisonerCompliancy: relocationAndInjuries.relocationCompliancy,
+    prisonerCompliant: relocationAndInjuries.prisonerCompliant,
```
andrewrlee: I don't think you can just change this? Won't this break for existing reports? 
Got to think of the data stored in json payloads in the DB.

server/views/formPages/incidentMacros.njk
```diff
@@ -36,7 +36,15 @@
 </div>
 {% endmacro %}
 
+
 {% macro radiosWithNestedRadios(question) %}
+
+{% if question.orientation === 'inline'  %}
+{% set radio_button_orientation = 'govuk-radios--inline'%}
+{% else %}
+{% set radio_button_orientation = 'govuk-radios' %}
```
andrewrlee: would have hoped that if this adhered to the standard GDS pattern, then we wouldn't have to use (and so adapt) our custom macro for this case..

server/views/formPages/incidentMacros.njk
```diff
@@ -36,7 +36,15 @@
 </div>
 {% endmacro %}
 
+
 {% macro radiosWithNestedRadios(question) %}
+
+{% if question.orientation === 'inline'  %}
+{% set radio_button_orientation = 'govuk-radios--inline'%}
```
andrewrlee: `{% set radio_button_orientation = 'govuk-radios--inline' if  question.orientation === 'inline' else 'govuk-radios' '%}`

server/utils/nunjucksSetup.js
```diff
@@ -71,4 +71,12 @@ module.exports = (app, path) => {
     }
     return value ? 'Yes' : 'No'
   })
+
+  njkEnv.addFilter('ifDefined', value => {
+    return value === undefined ? '' : ` - ${value}`
```
andrewrlee: this is too specific to have as a general macro and what it does is [surprising](https://wiki.c2.com/?PrincipleOfLeastAstonishment). 

All the other ways of formatting values appear inside ReportSummary, I think it would be better to be consistent and do this for this case.

server/routes/checkYourAnswers.test.js
```diff
@@ -121,6 +121,48 @@ describe('GET /check-your-answers', () => {
...
+      },
+    })
+
+    return request(app)
+      .get('/report/-35/check-your-answers')
+      .expect(200)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Was the prisoner compliant?')
+        expect(res.text).toContain('Yes')
```
andrewrlee: these tests are generally very good but this specific line is a bit off... its a big page and you can't really say that the `yes` is tied to that question!

server/views/formPages/incident/relocationAndInjuries.html
```diff
@@ -30,19 +30,43 @@ <h1 class="govuk-heading-xl mainHeading">{{ pageTitle }}</h1>
...
+        }],
+        errorMessage: errors | findError('prisonerCompliant')
+
+      },
+      followUpQuestion: {
+        text: "What was the type of relocation?",
+        name: "relocationType",
+        value: data.relocationType,
+        errorMessage: errors | findError('relocationType'),
+        options: [{
```
andrewrlee: We tend to store enumerated types in here: `/Users/andrewlee/code/use-of-force/server/config/types.js`.

you can then use the `toOptions` helper, e.g:

```    
   options: data.types.ControlAndRestraintPosition | list | extractAttr('value') | toOptions('value', 'label')
 ```
the idea being that we have encoded the 'Type' of the data explicitly rather than having the type implicitly defined in the html.

File: server/routes/model/reportSummary.js (hunk 1 above)
Comment (GurnankCheema) [reply server/routes/model/reportSummary.js]: you are quite right

File: server/views/formPages/incident/relocationAndInjuries.html (hunk 6 above)
Comment (GurnankCheema) [reply server/views/formPages/incident/relocationAndInjuries.html]: I tried to use the same pattern as the other components on the page. 
Have now changed to using types

File: server/routes/model/reportSummary.js (hunk 1 above)
Comment (GurnankCheema) [reply server/routes/model/reportSummary.js]: agreed

File: server/utils/nunjucksSetup.js (hunk 4 above)
Comment (GurnankCheema) [reply server/utils/nunjucksSetup.js]: done

File: server/views/formPages/incident/relocationAndInjuries.html (hunk 6 above)
Comment (GurnankCheema) [reply server/views/formPages/incident/relocationAndInjuries.html]: done

File: server/views/formPages/incidentMacros.njk (hunk 3 above)
Comment (GurnankCheema) [reply server/views/formPages/incidentMacros.njk]: ok

File: server/views/formPages/incidentMacros.njk (hunk 2 above)
Comment (GurnankCheema) [reply server/views/formPages/incidentMacros.njk]: done this way

File: server/routes/checkYourAnswers.test.js (hunk 5 above)
Comment (GurnankCheema) [reply server/routes/checkYourAnswers.test.js]: hmm.
I wanted to be able to test for the 'yes' if it was the only question on the page. 
So I have done this for the other sections
        incidentDetails: {},
        useOfForceDetails: {},
        evidence:{},
and the relocationAndInjuries object only has the one question answered
??

Comment (andrewrlee) [reply server/routes/checkYourAnswers.test.js]: It's not very easy to test at the moment.. some test framework provide css selector style ways of targeting elements. You could install [cheerio](https://github.com/cheeriojs/cheerio) as a dev dependency and try that if you want?

File: server/routes/model/reportSummary.js [hunk 7]
```diff
@@ -41,7 +42,10 @@ const createUseOfForceDetails = (details = {}) => {
 const createRelocation = (relocationAndInjuries = {}) => {
   return {
     prisonerRelocation: toLabel(RelocationLocation, relocationAndInjuries.prisonerRelocation),
-    prisonerCompliancy: relocationAndInjuries.relocationCompliancy,
+    relocationCompliancy: whenPresent(relocationAndInjuries.relocationCompliancy, value =>
+      value ? 'Yes' : `No - ${toLabel(RelocationType, relocationAndInjuries.relocationType)}`
```
andrewrlee: don't think this will work for old reports will it? when relocationType is null

server/routes/checkYourAnswers.test.js
GurnankCheema: ok. I 'll remove it for the time being.

server/services/reportStatusChecker.test.js
```diff
@@ -45,6 +45,7 @@ describe('statusCheck', () => {
       healthcareInvolved: true,
       prisonerRelocation: 'OWN_CELL',
       relocationCompliancy: false,
+      relocationType: 'relocation_to_vehicle',
```
andrewrlee: doesn't really matter but this should techincally be `VEHICLE`, would be generally good to refer to the constants you set up in Types. `Types.VEHICLE.value` or whatever

GurnankCheema: i see

PR #278, opened 2020-03-18
DCS-200 Adding new incident overview report

andrewrlee:
This returns summary information for all submitted Incidents that were submitted within a
specific month.

Waiting to here back whether we want to key by incident date instead but there are some
disadvantages that way based on getting incomplete reports at the end of the month.

Also moving to use new role based middleware rather than having custom authorisation within
the routes

### DIFF
--- a/server/services/reportingService.test.js
+++ b/server/services/reportingService.test.js
@@ -4,6 +4,7 @@ const serviceCreator = require('./reportingService')
 const reportingClient = {
   getMostOftenInvolvedStaff: jest.fn(),
   getMostOftenInvolvedPrisoners: jest.fn(),
+  getIncidentsOverview: jest.fn(),
 }
 
 const offenderService = {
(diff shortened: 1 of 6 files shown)


server/services/reportingService.test.js
```diff
@@ -68,7 +69,39 @@ Bella,10
...
 
-    expect(reportingClient.getMostOftenInvolvedPrisoners).toBeCalledWith('LEI', startDate, endDate)
+    expect(reportingClient.getMostOftenInvolvedPrisoners).toBeCalledWith('LEI', [startDate, endDate])
     expect(offenderService.getOffenderNames).toBeCalledWith('token-1', ['AAAA', 'BBBB', 'CCCC'])
   })
+
+  it('getIncidentsOverview', async () => {
+    reportingClient.getIncidentsOverview.mockReturnValue([
+      {
+        total: 'total-1',
```
bell-pa: It probably doesn't matter that you've used strings, but shouldn't each of the values be a number?

andrewrlee: yeah - it was just easier to provide a unique string than a unique number.. being a bit lazy. I can correct..

PR #282, opened 2020-03-20
DCS-330 Fixing cypress error

andrewrlee:
new [bug](https://github.com/cypress-io/cypress/issues/6788) in cypress which fails docker build.

Temporarily reverting cypress whilst its fixed.. have work on a separate branch to ensure we don't install cypress during the docker build


andrewrlee:
sdd

PR #283, opened 2020-03-20
DCS-416 upgrading dependencies

andrewrlee:
Some of these were quite far behind so would be good to do a bit of regression.
Cypress will probably break master docker build  as it now requires git to grab additional files.

We could do with upgrading the docker build file to not include cypress before merging


server/services/validation/errorDetailAdapter.js
```diff
@@ -30,13 +30,7 @@ const detailAdapterBuilder = firstFieldNameMap => {
...
 const extractFirstFieldNameMap = R.pipe(
   R.propOr({}, 'keys'),
-  R.map(
-    R.pipe(
-      R.prop('metas'),
-      R.mergeAll,
-      R.prop('firstFieldName')
-    )
-  ),
+  R.map(R.pipe(R.prop('metas'), R.mergeAll, R.prop('firstFieldName'))),
```
andrewrlee: seems like prettier is a bit broken here.. couldn't work out how to fix to use existing behaviour

PR #288, opened 2020-03-29
DCS-429 Moving to deploy compiled application

andrewrlee:
Moving to deploy typescript compiled application as part of the
docker image.

Also updating dev scripts to run against compiled location.
When debugging, breakpoints will have to be set in the compiled code
rather than the typescript/js source - all resources will be copied across
automatically in :dev mode

The docker multistage build reduces image size from 850MB to just under 300MB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,11 @@
-FROM node:10-buster-slim
-LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
+# Build stage 1.
+# Create base image with python (required for node-gyp)
 ARG BUILD_NUMBER
 ARG GIT_REF
 
-RUN apt-get update && \
-    apt-get upgrade -y && \
-    apt-get install -y make python curl wget && \
-    apt-get autoremove -y && \
-    rm -rf /var/lib/apt/lists/*
+FROM node:10-buster-slim as base
+
+LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
 ENV TZ=Europe/London
 RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezone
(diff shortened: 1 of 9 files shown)


server/app.js
```diff
@@ -128,17 +129,17 @@ module.exports = function createApp({
...
-    '../node_modules/@ministryofjustice/frontend/',
+    '/assets',
+    '/assets/stylesheets',
+    '/assets/js',
+    `/node_modules/govuk-frontend/govuk/assets`,
+    `/node_modules/govuk-frontend`,
+    `/node_modules/@ministryofjustice/frontend/`,
   ].forEach(dir => {
-    app.use('/assets', express.static(path.join(__dirname, dir), cacheControl))
+    app.use('/assets', express.static(path.join(process.cwd(), dir), cacheControl))
```
andrewrlee: need to work both when deployed and locally
`__dirname` shows location of this file whereas cwd is the directory where node is run from.

Using `__dirname`, locally when code is compiled to a `dist` folder, node_modules is `../../` away whereas when running in docker its `../`

With `cwd` it's alway at the top level

package.json
```diff
@@ -5,21 +5,24 @@
...
   "scripts": {
-    "start": "node server.js",
-    "start:dev": "npm run build && DEBUG=gov-starter-server* DB_PORT=5433 nodemon server.js | bunyan -o short",
-    "start-feature": "NOTIFY_ENABLED=false PORT=3007 NOMIS_AUTH_URL=http://localhost:9091/auth ELITE2API_ENDPOINT_URL=http://localhost:9091 DB_NAME=use-of-force-int DB_PORT=5432 NODE_ENV=development node $NODE_DEBUG_OPTION server.js | bunyan -o short",
-    "start-feature:dev": "NOTIFY_ENABLED=false PORT=3007 NOMIS_AUTH_URL=http://localhost:9091/auth ELITE2API_ENDPOINT_URL=http://localhost:9091 DB_NAME=use-of-force-int DB_PORT=5432 NODE_ENV=development nodemon $NODE_DEBUG_OPTION server.js | bunyan -o short",
-    "build": "npm run css-build",
+    "watch-views": "nodemon --watch server/views -e html,njk -x cp -R server/views/ dist/server/views/",
+    "watch-ts": "tsc -w",
+    "watch-node": "DEBUG=gov-starter-server* DB_PORT=5433 nodemon --watch dist/ dist/server.js | bunyan -o short",
+    "start:dev": "npm run build && concurrently -k -p \"[{name}]\" -n \"Views,TypeScript,Node\" -c \"yellow.bold,cyan.bold,green.bold\" \"npm run watch-views\" \"npm run watch-ts\" \"npm run watch-node\"",
```
andrewrlee: Will look at extending this to include SASS in a follow up

Dockerfile
```diff
@@ -1,40 +1,73 @@
...
-COPY --chown=appuser:appgroup . .
+COPY . .
 
 ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 ENV GIT_REF ${GIT_REF:-dummy}
 
-RUN npm ci && \
-    npm run build && \
+# Remove cypress as it takes ~1 minute to download and install
+RUN sed -ie 's/"cypress.*//g' package.json
```
andrewrlee: shame npm doesn't distinguish between build and test dependencies

Dockerfile
```diff
@@ -1,40 +1,73 @@
...
+        /app/build-info.json \
+        ./
+
+COPY --from=build --chown=appuser:appgroup \
+        /app/assets ./assets
+
+COPY --from=build --chown=appuser:appgroup \
+        /app/server/views ./server/views
+
+RUN npm ci --only=production
```
andrewrlee: This ends up doing npm install twice..

an alternative would be copying over `node_modules` and doing `npm prune --production` but this seems to take longer and also results in slightly larger docker images (though has the benefit of not downloading the internet again)

Dockerfile
```diff
@@ -1,40 +1,73 @@
-FROM node:10-buster-slim
-LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
+# Build stage 1.
+# Create base image with python (required for node-gyp)
 ARG BUILD_NUMBER
 ARG GIT_REF
```
andrewrlee: defining global args to be used by individual stages

Dockerfile
```diff
@@ -1,40 +1,73 @@
...
+    apt-get upgrade -y && \
+    apt-get install -y make python
+
+# Build stage 2.
+# This stage builds our assets.
+FROM base as build
+ARG BUILD_NUMBER
+ARG GIT_REF
+
+RUN apt-get install -y curl wget
```
andrewrlee: only install curl and wget on base image rather than prod image

Dockerfile
```diff
@@ -1,40 +1,73 @@
...
+        /app/dist \
+        /app/root.cert \
+        /app/build-info.json \
+        ./
+
+COPY --from=build --chown=appuser:appgroup \
+        /app/assets ./assets
+
+COPY --from=build --chown=appuser:appgroup \
+        /app/server/views ./server/views
```
andrewrlee: have to do multiple `COPY` to preserve paths

package.json
```diff
@@ -77,12 +80,10 @@
   "lint-staged": {
     "*.{js,css}": [
       "prettier --write",
-      "eslint --fix",
-      "git add"
```
andrewrlee: `lint-staged` automatically does `git add` in later versions

PR #290, opened 2020-03-31
DCS-429 Adding initial typescript file

andrewrlee:
Also removing node-sass-middleware as we never really used this as it makes node-sass a prod dependency (which has lots of security vulns)

I think we're not currently linting typescript files but prettier works on them


package.json
```diff
@@ -155,8 +156,11 @@
     "mocha": "^6.2.2",
     "mocha-junit-reporter": "^1.23.1",
     "nock": "^11.4.0",
+    "node-sass": "^4.13.1",
+    "nodemon": "^2.0.2",
```
andrewrlee: adding nodemon as a dev dependency as the scripts require it (and now node-sass is required as not pulled in as part of the sass middleware)

server/config/forms/statementValidation.test.js
```diff
@@ -1,3 +1,8 @@
+jest.mock('moment', () => () => ({
```
andrewrlee: jest/babel previously hoisted this up to the beginning of the file - we need to do it manually now as using ts-jest rather than babel for compilation of tests

server/index.js
```diff
@@ -14,7 +14,7 @@ const { authClientBuilder, systemToken } = require('./data/authClientBuilder')
 
 const createSignInService = require('./authentication/signInService')
 
-const createHeatmapBuilder = require('./services/heatmapBuilder')
+const createHeatmapBuilder = require('./services/heatmapBuilder').default
```
andrewrlee: now we also export types we need to explicitly specify the `default` export

PR #291, opened 2020-04-01
DCS-429 Fixing missing request library preventing app from starting up in dev

andrewrlee:
I think this might have been included in node-sass-middleware and only required by app insights - so it got all the way until attempting to deploy to dev before failing


andrewrlee:
> I don't understand why this has to be an explicit dependency if it is already a transitive dependency of another module.
> Nevertheless: LGTM
It is a peer dependency of `request-promise` so has to be installed separately: 
https://github.com/request/request-promise/issues/142

PR #298, opened 2020-04-05
DCS-433 moving to use DPS orb

andrewrlee:


.circleci/config.yml
```diff
@@ -166,90 +108,85 @@ jobs:
...
       - build:
-          filters:
-            tags:
-              ignore: /.*/
+        filters:
+          tags:
+            ignore: /.*/
       - test:
           requires:
             - build
```
petergphillips, line 137: I've also been renaming `build` and the `builder` executor to `validate` and `validator` since they don't actually do any building any more

andrewrlee: they kind of do in use of force - they build the compiled typescript which the integration tests are then run against

petergphillips: interesting, didn't realise that :+1:

.circleci/config.yml
```diff
@@ -1,68 +1,16 @@
...
-          condition: <<parameters.slack_notification>>
-          steps:
-            - slack/notify:
-                message: "*${APP}* version:*${APP_VERSION}* deployed to << parameters.env >>"
-                include_project_field: false
-                include_job_number_field: false
+parameters:
+  image_name:
+    type: string
+    default: "mojdigitalstudio/use-of-force"
```
petergphillips, line 5: hoping to switch to quay as well - see quay lines in https://dsdmoj.atlassian.net/wiki/spaces/NOM/pages/1747943435/Creating+a+new+microservice

PR #299, opened 2020-04-06
DCS-433 moving to quay

andrewrlee:


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.1'
 services:
 
   oauth-server:
-    image: mojdigitalstudio/nomis-oauth2-server:latest
```
petergphillips, line 5: shouldn't be changed

PR #305, opened 2020-04-06
DCS-199 Incidents by race/ethnic group

bell-pa:
Very similar to DCS 198 'Incidents by religion'

DCS-198 Counted incidents using a classification derived from the `religionCode` property of the prisoner's details. (Prisoner details are retrieved from elite2-api for each offender involved in an incident).

DCS-199 Counts incidents using a classification derived from the `ethnicityCode` property of the prisoner's details.

Refactored to extract common code.  

Switched key javascript files to Typescript and introduced some type information.  This is meant to clarify how Aggregator implementations are defined, created and used by reportingService.ts.  There are two, one each for DCS-198 and DCS-199.

Also tweaked the way Application Insights is configured to ensure that the configuration happens early, but in a way that is tolerated by lint, ts, prettier etc.  This needs improving, but that might be easier when more of the code has been migrated to Typescript.


server/index.js
```diff
@@ -1,6 +1,6 @@
-// Do appinsights first as it does some magic instrumentation work, i.e. it affects other 'require's
-// In particular, applicationinsights automatically collects bunyan logs
-const appInsightsClient = require('./utils/azure-appinsights')()
+import buildAppInsightsClient from './utils/azure-appinsights'
+import createOffenderService from './services/offenderService'
+import createReportingService from './services/reportingService'
```
andrewrlee, line 3: 👍 sweet, I thought you could mix the 2 import styles but got into a bit of a pickle with linting rules

server/types/uof.d.ts
```diff
@@ -1,34 +1,59 @@
...
-    activeCaseLoadId: string
-    accountStatus: String
-    active: Boolean
-    caseLoadId: String
-    description: String
-    type: String
-    caseloadFunction: String
-    currentlyActive: Boolean
-    displayname: String
+  staffId: number
```
andrewrlee, line 2: 👍

server/utils/azure-appinsights.ts
```diff
@@ -0,0 +1,39 @@
+import appInsights from 'applicationinsights'
+import applicationVersion from '../application-version'
+
+if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY) {
```
andrewrlee, line 4: won't this get run for the job too? meaning application insights starts twice?

bell-pa: Isn't the job run in a separate process?  If not then I've made a mistake - or I misunderstand when code is loaded/run - or both!

PR #308, opened 2020-04-17
DCS-196 Incidents by age group

bell-pa:


server/services/offenderService.ts
```diff
@@ -43,31 +43,32 @@ export default function createOffenderService(elite2ClientBuilder) {
...
     const offenders = await elite2ClientBuilder(token).getOffenders(uniqueNos)
+
     return offenders.reduce((rv, offender) => ({ ...rv, [offender.offenderNo]: fullName(offender) }), {})
   }
 
-  const getLocation = (token, locationId) => {
+  const getLocation = (token: string, locationId: string): Promise<PrisonLocation> => {
     if (!locationId) {
-      return {}
+      return Promise.resolve({})
```
andrewrlee, line 65: meh.. would have thought that you could just return `{}` as returning from an async function will wrap the object in a Future.

server/types/uof.d.ts
```diff
@@ -24,18 +40,53 @@ type GetUsersResults = {
...
+}
+
 export interface UserService {
   /** get current user */
   getUser: (token: string) => Promise<User>
   /** Get details for users along with email address */
   getUsers: (token: string, usernames: string[]) => Promise<GetUsersResults[]>
 }
 
+export interface OffenderService {
```
andrewrlee, line 64: Think it would probably be good to get the services/clients themselves to export their own types. 
So this could probably live in OffenderService. Not too sure but could even derive it based on the service object, or maybe turn the service object into a class?

(As reporting client is directly returned (and it's not a factory) it's not strictly needed to define an interface for it merely for type safety - the type can just be imported directly)

server/services/offenderService.ts
bell-pa: Oh yes!

server/types/uof.d.ts
bell-pa: Sounds good to me. 
I couldn't decide if clients should export their types, or if the 'domain' should define that are then implemented by the clients.  If the types and interfaces are part of the domain then the implementations should depend upon those rather then the domain depending upon the implementations...

PR #310, opened 2020-04-24
Dcs 424 sort prisoner and other cell

GurnankCheema:
The UOF report incident-details page has a select input for Incident location.
This PR is to ensure that should there be any locations called Prisoner's Cell or Other cell, they are at the top of the select options.
Any subsequent options will be in alpha order

### DIFF
--- a/server/services/offenderService.test.ts
+++ b/server/services/offenderService.test.ts
@@ -83,6 +83,80 @@ describe('getOffenders', () => {
       expect(elite2Client.getLocations).toBeCalledWith('WRI')
     })
 
+    it('should sort retrieved locations with top 2 primary locations at the begining', async () => {
+      elite2Client.getLocations.mockReturnValue([
+        { id: 2, userDescription: 'place 2' },
+        { id: 'o', userDescription: 'Other cell' },
+        { id: 3, userDescription: 'place 3' },
+        { id: 'p', userDescription: "Prisoner's cell" },
+        { id: 1, userDescription: 'place 1' },
+        { id: 4, userDescription: 'place 4' },
+      ])
+
+      const result = await service.getIncidentLocations(token, 'WRI')
+
+      expect(result).toEqual([
+        { id: 'p', userDescription: "Prisoner's cell" },
+        { id: 'o', userDescription: 'Other cell' },
+        { id: 1, userDescription: 'place 1' },
+        { id: 2, userDescription: 'place 2' },
+        { id: 3, userDescription: 'place 3' },
+        { id: 4, userDescription: 'place 4' },
+      ])
+    })
+
+    it('should sort retrieved locations with only 1 primary location at the begining', async () => {
+      elite2Client.getLocations.mockReturnValue([
+        { id: 2, userDescription: 'place 2' },
+        { id: 3, userDescription: 'place 3' },
+        { id: 'p', userDescription: "Prisoner's cell" },
+        { id: 1, userDescription: 'place 1' },
+        { id: 4, userDescription: 'place 4' },
+      ])
+
+      const result = await service.getIncidentLocations(token, 'WRI')
+
+      expect(result).toEqual([
+        { id: 'p', userDescription: "Prisoner's cell" },
+        { id: 1, userDescription: 'place 1' },
+        { id: 2, userDescription: 'place 2' },
+        { id: 3, userDescription: 'place 3' },
+        { id: 4, userDescription: 'place 4' },
+      ])
+    })
+
+    it('should sort retrieved locations with zero primary locations at the begining', async () => {
+      elite2Client.getLocations.mockReturnValue([
+        { id: 2, userDescription: 'place 2' },
+        { id: 3, userDescription: 'place 3' },
+        { id: 1, userDescription: 'place 1' },
+        { id: 4, userDescription: 'place 4' },
+      ])
+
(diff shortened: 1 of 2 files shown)


server/services/offenderService.ts
```diff
@@ -71,8 +81,25 @@ export default function createOffenderService(elite2ClientBuilder): OffenderServ
   const getIncidentLocations = async (token: string, agencyId: AgencyId): Promise<PrisonLocation[]> => {
     try {
       const elite2Client = elite2ClientBuilder(token)
+      const incidentLocations = await elite2Client.getLocations(agencyId)
+
+      const primaryLocations = incidentLocations.filter(
+        loc =>
+          loc.userDescription.toUpperCase() === "PRISONER'S CELL" || loc.userDescription.toUpperCase() === 'OTHER CELL'
```
brightonsbox, line 88: might be a bit cleaner to specify an array of locations considered 'primary'

so this would become something like:

```
const primaryLocationOptions = ["PRISONER'S CELL", 'OTHER CELL']
const primaryLocations = incidentLocations.filter(
  loc => primaryLocationOptions.includes(loc.userDescription.toUpperCase()))
const remainingLocations = incidentLocations.filter(
  loc => !primaryLocationOptions.includes(loc.userDescription.toUpperCase())
)
.sort(compare)

File: server/services/offenderService.test.ts [hunk 2]
```diff
@@ -83,6 +83,80 @@ describe('getOffenders', () => {
       expect(elite2Client.getLocations).toBeCalledWith('WRI')
     })
 
+    it('should sort retrieved locations with top 2 primary locations at the begining', async () => {
+      elite2Client.getLocations.mockReturnValue([
+        { id: 2, userDescription: 'place 2' },
+        { id: 'o', userDescription: 'Other cell' },
```
andrewrlee, line 89: the ids that are returned from elite are numeric, doesn't affect your tests too much but would be good to keep them consistent with the real types

server/services/offenderService.ts
```diff
@@ -2,6 +2,16 @@ import logger from '../../log'
...
 import { AgencyId, OffenderService, PrisonerDetail, PrisonLocation } from '../types/uof'
 
+const compare = (a, b): number => {
+  let comparison = 0
+  if (a.userDescription.toUpperCase() > b.userDescription.toUpperCase()) {
+    comparison = 1
+  } else if (a.userDescription.toUpperCase() < b.userDescription.toUpperCase()) {
+    comparison = -1
+  }
+  return comparison
```
andrewrlee, line 12: you shouldn't need to create your own comparison function: you can use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

server/services/offenderService.ts
brightonsbox: You could use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) instead of the compare method:
```
.sort((a,b) => a.userDescription.localeCompare(b.userDescription, 'en', {sensitivity: 'base'})
```

server/services/offenderService.ts
```diff
@@ -71,8 +81,25 @@ export default function createOffenderService(elite2ClientBuilder): OffenderServ
...
+        .filter(
+          loc =>
+            loc.userDescription.toUpperCase() !== "PRISONER'S CELL" &&
+            loc.userDescription.toUpperCase() !== 'OTHER CELL'
+        )
+        .sort(compare)
+
+      if (primaryLocations.length === 2 && primaryLocations[0].userDescription.toUpperCase() !== "PRISONER'S CELL") {
+        primaryLocations.reverse()
+      }
```
andrewrlee, line 100: Having to handle this edge case isn't great.. would be good to avoid hardcoding a check of the first element and then having to mutate the array.  

An alternative would be to be more explicit with the finding and placing of the special elements:
```js
const  remainingLocations = ...

const prisonersCell = incidentLocations.find(...
const otherCell = incidentLocations.find(...

return [
...(prisonersCell ? [prisonersCell] : []), 
...(othersCell ? [othersCell] : []), 
...remainingIncidents,
]

File: server/services/offenderService.ts (hunk 1 above)
Comment (brightonsbox) [reply server/services/offenderService.ts:88]: Or instead of pulling out the separate lists, could do the following:

```
const compare = (a, b): number => {
  if (a.userDescription.toUpperCase() === "PRISONER'S CELL") return -1;
  if (a.userDescription.toUpperCase() === "OTHER CELL") {
    return b.userDescription.toUpperCase() === "PRISONER'S CELL" ? 1 : -1;
  }
  return a.userDescription.localeCompare(b.userDescription, 'en', {sensitivity: 'base'})
}
return incidentLocations.sort(compare)
```

server/services/offenderService.ts
GurnankCheema: had this convo before... doh!

server/services/offenderService.ts
GurnankCheema: used the locale compare.

server/services/offenderService.ts
GurnankCheema: ordering by placing items in order in return array is much simpler! 
done.

PR #317, opened 2020-05-07
DCS-465 Adding frontend for prison search page

andrewrlee:
This is just the frontend - it returns stubbed data if searched with any data. 
Supports sorting of results on name too.

<kbd>
Screenshot 2020-05-07 at 14 37 36
</kbd>

On a post, we currently pass search terms via a redirect and the query string back to the view page and will optionally perform a "search" if those parameters exist. 
Weighed up passing terms by flash state (difficult to test) or render on post

There isn't currently any validation to enforce only searching on prison number or by name and prison but still discussing the details of how the search will actually work

 Also still need to load prison list but Gurnank has got code that performs this query.


assets/js/sortable-table.js
```diff
@@ -0,0 +1,161 @@
+// This is adapted from https://github.com/LJWatson/sortable-tables
```
andrewrlee, line 1: this is really based on the prison staff hub version

andrewrlee, line 1: this is actually taken from prison staff hub

PR #318, opened 2020-05-12
DCS-465 Hooking up search backend

andrewrlee:
Still need to consider validation and need to relax security around creation of reports.
Currently we have a check to ensure that the user should be able to see the prisoner before
allowing them to create a report.


helm_deploy/values-prod.yaml
```diff
@@ -20,6 +20,7 @@ env:
   NOMIS_AUTH_URL: https://gateway.prod.nomis-api.service.hmpps.dsd.io/auth
   NOMIS_OAUTH_PUBLIC_KEY: LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUlJQklqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FROEFNSUlCQ2dLQ0FRRUFwSkdtYWp6SW9POVdSY2owM2VTbQpFeWJTYnNJaEdkNkRWVDRxUWFSRTZwN3RIN1ppSklvRWtzczRkOFphMGhuYWJwa0JqWTYwS3NnMW5teXpoZVlSCithWXVtc0gxbm1JcmJ4ZCtqV043T3A2V1pteWRBU3JMek1XWVdBYlNHOHBXMXkxY1RwdkRaMTlBSGJTYTRRL2UKSnRlcDFscCsyMnVHMW5kTVRwclRkNjRLTkcyZnRwcmV3aFExeE5xdmtUWjZoQnVRRVJKMzFhUXBkRThNQTFxSwpDWnI3VDJxTk1BTWNkTlpia0FqRjVLOXE0alZ0QkY3WXFPeXJ0T1JmTldqYUJoWHA5TlN0YWJCTXBURTdQNm8vClhJaU9rcmltNjlzaG1OdDhVRlU3WUlJN0ZvejgzM29HZnk2dmtDRjRxcGpIZ0RNc21STjc1UXRrRmRtaU8xaDgKcHdJREFRQUIKLS0tLS1FTkQgUFVCTElDIEtFWS0tLS0tCg==
   ELITE2_API_URL: https://gateway.prod.nomis-api.service.hmpps.dsd.io/elite2api
+  PRISONER_SEARCH_ENDPOINT_URL: https://prisoner-offender-search.prison.service.justice.gov.uk
```
andrewrlee, line 23: don't think this exists yet

server/data/elite2ClientBuilder.js
```diff
@@ -1,161 +1,53 @@
...
         errorLogger: error =>
           error.status === 404
             ? logger.info(`No offender image available for: ${bookingId}`)
-            : defaultErrorLogger(error),
+            : logger.warn(error, `Error calling elite2`),
       })
     },
   }
 }
-
```
andrewrlee, line 73: extracted out the guts of this so can reuse in new search client

server/data/prisonerSearchClient.ts
```diff
@@ -0,0 +1,34 @@
...
+const restClientBuilder = createRestClientBuilder('prisonerSearch', config.apis.prisonerSearch)
+
+export type SearchForm = {
+  prisonNumber?: string
+  firstName?: string
+  lastName?: string
+  agencyId?: string
+}
+
+export type PrisonerSearchClientBuilder = new (token: string) => PrisonerSearchClient
```
andrewrlee, line 13: This type matches the constructor of the client. This allows us to pass the class as an argument in the service:

`new PrisonSearchService(PrisonSearchClient, ... )`

Where the constructor of PrisonSearchService looks like:

` constructor(PrisonerSearchClient: PrisonerSearchClientBuilder,...`

We can also mock it out relatively easy with jest.

helm_deploy/values-prod.yaml
Mjwillis: Correct!

PR #321, opened 2020-05-12
DCS 285 user can change prison of incident

GurnankCheema:
The user can now record a prison different to the one the prisoner is currently at. This case might arise if a prisoner has been moved Prison A to Prison B and staff at Prison A need to raise a UoF incident. 
Pages impacted 
http://localhost:3000/report/-19/incident-details
http://localhost:3000/report/-19/change-prison
http://localhost:3000/report/-19/check-your-answers

A user will begin the report journey at http://localhost:3000/report/-19/incident-details
The prison showing on this page will be the one where the prisoner is currently residing. The details for this are obtained from `offenderService.getOffenderDetails()`

It is necessary to display this as the as the default prison 

The change of prison is done in the 'change-prison' page 

At present however before a change can be recorded, a report must first have been initiated by completing and saving either of the 4 main pages of the report. 
Allowing the report to be initiated as soon as the user selects the 'change' link to go to the change-prison page will be done in due course. 

If a user changes a prison then on returning to the incident-details page, they may need to update the internal location dropdown selection.

It is assumed elite2 will be updated so as to provide prison details from the corresponding agencyId and also provide the corresponding internal locations.


server/config/forms/changePrisonForm.js
```diff
@@ -0,0 +1,12 @@
+const { joi, validations } = require('./validations')
+const { buildValidationSpec } = require('../../services/validation')
+
+const { requiredStringMsg } = validations
+
+const transientSchema = joi.object({
+  agencyId: requiredStringMsg('What prison did the use of force take place in?'),
```
andrewrlee, line 7: I think we talked about this but wouldn't bother to handle the extra complexity of using the form processor validation stuff. It's only one field. 

You might has well just check it in the route.

server/data/elite2ClientBuilder.js
```diff
@@ -49,5 +49,16 @@ module.exports = token => {
             : logger.warn(error, `Error calling elite2`),
       })
     },
+    getActiveAgenciesByType(type) {
```
andrewrlee, line 52: I added a `getPrisons` method to this class which you can re-use instead of this. 

I aimed for encapsulating the quirks of elite2 into just the client - no need to spread the knowledge that elite stores prisons as "agencies" of a type of `INST` anywhere but in here

server/data/elite2ClientBuilder.js
```diff
@@ -49,5 +49,16 @@ module.exports = token => {
...
+      return restClient.get({
+        path: `/api/agencies/type/${type}`,
+      })
+    },
+
+    getPrisonById(agencyId) {
```
andrewrlee, line 58: We're a bit inconsistent with what we call the id of prisons - in the future we can look at normalising this.

server/data/incidentClient.js
```diff
@@ -46,6 +46,17 @@ const updateDraftReport = (reportId, incidentDate, formResponse) => {
 const maxSequenceForBooking =
   '(select max(r2.sequence_no) from report r2 where r2.booking_id = r.booking_id and user_id = r.user_id)'
 
+const updateAgencyId = (agencyId, username, bookingId) => {
+  return nonTransactionalClient.query({
+    text: `update v_report r
+                set agency_id = COALESCE($1,   r.agency_id)
```
andrewrlee, line 52: obviously no actual need for the COALESCE as validation wouldn't allow ever setting it to null but no harm

server/routes/checkYourAnswers.js
```diff
@@ -38,8 +47,11 @@ module.exports = function CheckAnswerRoutes({ reportService, offenderService, in
       ]
 
       const data = reportSummary(form, offenderDetail, locationDescription, involvedStaff, incidentDate)
-
-      return res.render('pages/check-your-answers', { data, bookingId })
+      const prison = await locationService.getPrisonById(token, prisonId).catch(err => {
+        logger.warn(err, 'Prison details not found in elite-2')
+        return { description: getIn(['assignedLivingUnit', 'agencyName'], offenderDetail) }
```
andrewrlee, line 52: I don't think you should be handling the error here. 

This should never happen and we shouldn't handle things thats will reasonably never happen.
 
If it does happen this could potentially show the wrong prison which is probably worse than blowing up.

andrewrlee: also would suggest not using Ramda unless strictly necessary as it obfuscates intent around null handling and removes any type safety

server/data/elite2ClientBuilder.js
```diff
@@ -49,5 +49,16 @@ module.exports = token => {
...
+      return restClient.get({
+        path: `/api/agencies/prison/${agencyId}`,
```
andrewrlee, line 60: Think it might be better to use `/api/agencies/{agencyId}?activeOnly=false`

It returns the least amount of data that you require - for instance, doesn't rely on contact details and is a bit more explicit around that you would want to potentially return inactive prisons too.

andrewrlee: In fact, its a good idea to check to see what the API will actually return - looking at the elite2 codebase.

It seems in this case the path of the url suggests it's returning prisons but its really all about serving a use case around retrieving contact details - as such it doesn't make sense to return inactive prisons if you are only interested in contact details. 

If you track through the api call, you'll see it eventually leads down to a query called `FIND_PRISON_ADDRESSES_PHONE_NUMBERS` and this explicitly hard codes the active flag:   
`WHERE al.ACTIVE_FLAG = 'Y'`

server/routes/createReport.js
```diff
@@ -117,13 +124,21 @@ module.exports = function NewIncidentRoutes({ reportService, offenderService, in
     const involvedStaff =
       (input && input.involvedStaff) || (formId && (await involvedStaffService.getDraftInvolvedStaff(formId))) || []
 
+    const prison = await locationService.getPrisonById(token, agencyId).catch(err => {
+      logger.warn(err, 'Prison details not found in elite-2')
+      return { description: getIn(['assignedLivingUnit', 'agencyName'], offenderDetail) }
```
andrewrlee, line 129: similarly - don't think there is any point handling this.

server/routes/locations.js
```diff
@@ -0,0 +1,44 @@
+const { isNilOrEmpty } = require('../utils/utils')
```
andrewrlee, line 1: I don't think it makes sense to group routes by the type of entity that they deal with. We tend to group routes around use cases. So in this case it would be better to call this `changePrison` rather than `locations`

server/routes/locations.js
```diff
@@ -0,0 +1,44 @@
...
+const { processInput } = require('../services/validation')
+const { full } = require('../config/incident')
+
+module.exports = ({ locationService }) => {
+  // get list of all prisons
```
andrewrlee, line 6: No point having comments if they don't add any context as to why a decision was made or explaining complicated or surprising code that you can't express intent by rewriting/simplifying the code. 

Generally a well named method can replaces the need for a comment. 

In this case renaming the route along the line of use case `changePrison` means you don't really need the comment at all. If that code is executed then you are `"submit"ing a change of prison`

server/routes/locations.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  // submit new prison id
+  const submit = (formName, editMode) => async (req, res) => {
+    const { bookingId } = req.params
+    const { agencyId } = req.body
+    const { username } = res.locals.user
+    const saveAndContinue = req.body.submit === 'save-and-continue'
+
+    if (saveAndContinue) {
```
andrewrlee, line 22: so you can get rid of this `processInput` thing and just have:
``` 
  if (!agencyId) {
        req.flash('errors', ...)
        return res.redirect(req.originalUrl)
    }
```

server/routes/locations.test.js
```diff
@@ -0,0 +1,38 @@
...
+  jest.resetAllMocks()
+})
+
+describe('GET /change-prison', () => {
+  test('should display page title', () => {
+    return request(app)
+      .get(`/report/-19/change-prison`)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('What prison did the use of force take place in?')
```
andrewrlee, line 25: might want to check that a prisoned returned by change prison is present as a select option

server/routes/locations.test.js
```diff
@@ -0,0 +1,38 @@
...
+      })
+  })
+
+  test('should contain the Continue and Cancel buttons', () => {
+    return request(app)
+      .get(`/report/-19/change-prison`)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Continue')
```
andrewrlee, line 34: not sure this explicitly requires a separate test? You would want a test for submitting - to ensure the passed agency id is saved in the service

server/routes/searchForPrisoner.js
```diff
@@ -9,7 +9,7 @@ module.exports = function SearchForPrisonerRoutes({ prisonerSearchService }) {
   return {
     view: async (req, res) => {
       const { form, openDetails } = extractSearchForm(req.query)
-      const results = await await prisonerSearchService.search(form)
```
andrewrlee, line 12: thanks!

server/services/locationService.test.ts
```diff
@@ -0,0 +1,92 @@
...
+})
+
+afterEach(() => {
+  elite2Client.getActiveAgenciesByType.mockReset()
+  elite2Client.getPrisonById.mockReset()
+})
+
+describe('locationService', () => {
+  describe('getActiveAgenciesByType', () => {
+    it('should return no prison', async () => {
```
andrewrlee, line 28: `one prison` ? You don't strictly need this test- there isn't any particular logic around handling one prison

server/services/locationService.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      elite2Client.getPrisonById.mockReturnValue(prison)
+      const result = await locationService.getPrisonById(token, 'MDI')
+      expect(result).toEqual(prison)
+    })
+  })
+
+  describe('updateAgencyId', () => {
+    it('incidentClient.updateAgencyId should be called', async () => {
+      await locationService.updateAgencyId('x', 'y', 'z')
```
andrewrlee, line 88: good to use real variables here - tests are supposed to provide documentation. This doesn't give any indication as to how this method would be called in real life

server/routes/locations.js
```diff
@@ -0,0 +1,44 @@
+const { isNilOrEmpty } = require('../utils/utils')
+const { processInput } = require('../services/validation')
+const { full } = require('../config/incident')
+
+module.exports = ({ locationService }) => {
+  // get list of all prisons
+  const viewPrisons = editMode => async (req, res) => {
+    const { token } = res.locals.user
+    const prisons = await locationService.getActiveAgenciesByType(token, 'INST')
```
andrewrlee, line 9: Might be safer to request tokens as a system to system request rather than as a specific user.
if you look to see how `systemToken` is used, where we pass the user's username.

server/views/pages/reportDetailMacro.njk
```diff
@@ -64,6 +64,16 @@
               dataValue: data.incidentDetails.incidentDate | formatDate('HH:mm') 
       })
     }}
+    {% if prison %}
```
andrewrlee, line 67: is this `if` required? I think we would always want to show the prison?

andrewrlee: Think we need to load the prison when displaying "Your Report" and "View Report" too

server/config/forms/changePrisonForm.js
GurnankCheema: done. 
Validation removed from form processor and done directly in routes/location.js instead.
I do remember the conversation but thought it was primarily about whether we should have validation at all. When it was decided to keep the validation, I then assumed that  it should be done like every other input in the app, i.e via Joi.

server/data/elite2ClientBuilder.js
GurnankCheema: Done.
swapped over to use getPrisons

server/data/elite2ClientBuilder.js
GurnankCheema: Yes, I wasn't sure which name to use because in the report and elite2 its referred to as agencyId.

server/data/incidentClient.js
GurnankCheema: ok

server/routes/checkYourAnswers.js
GurnankCheema: Done. I thought about not having the error checking but included it as a fallback.

server/routes/locations.js
GurnankCheema: ok

server/routes/locations.js
GurnankCheema: done

server/routes/locations.test.js
GurnankCheema: done
new test added

server/routes/locations.test.js
GurnankCheema: I wanted to check that all the components that should be on a page were actually on there. 
Test removed. 
Will do cypress integration test for submission.

server/routes/searchForPrisoner.js
GurnankCheema: np

server/services/locationService.test.ts
GurnankCheema: done

server/services/locationService.test.ts
GurnankCheema: done

server/routes/locations.js
GurnankCheema: ok

server/views/pages/reportDetailMacro.njk
GurnankCheema: done

server/routes/createReport.js
GurnankCheema: I do think this needs exception-handling because when the user enters the incident-details page for the first time and hasn't yet saved the page, there won't be a value for agencyId. So calling locationService.getPrisonById(token, prisonId) will cause an exception. 
And in this instance, we need to display the actual prison the prisoner is currently at via offenderService.getOffenderDetails(token, bookingId)

server/routes/locations.js
GurnankCheema: Done
makes sense

server/data/elite2ClientBuilder.js
GurnankCheema: hmmm.... have changed the url to include ?activeOnly=false but it hasn't made a difference in terms of returning the prison details from its id ...

GurnankCheema: done

integration-tests/integration/reportPages/change-prison.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  it('Can edit prison prior to reaching check-your-answers page', () => {
+    completeIncidentDetails()
+
+    const incidentDetailsPage = IncidentDetailsPage.verifyOnPage()
+
+    incidentDetailsPage.clickChangePrison()
+
+    const changePrisonPage = ChangePrisonPage.verifyOnPage()
+
+    changePrisonPage.changePrisonToLeeds()
```
andrewrlee, line 50: would parameterise this, like we do with location below: `    incidentDetailsPage.location().select('Asso A Wing')`

integration-tests/integration/reportPages/change-prison.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  const fillFormAndSave = () => {
+    const reportUseOfForcePage = ReportUseOfForcePage.visit(bookingId)
+    const incidentDetailsPage = reportUseOfForcePage.startNewForm()
+    incidentDetailsPage.offenderName().contains('Norman Smith')
+    incidentDetailsPage.location().select('Asso A Wing')
+    incidentDetailsPage.forceType.check('true')
+    incidentDetailsPage.save()
+  }
+
+  const completeIncidentDetails = () => {
```
andrewrlee, line 35: might as well merge these two methods; `fillFormAndSave`, `completeIncidentDetails`- you aren't using either independently are you

integration-tests/integration/reportPages/change-prison.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    incidentDetailsPage.prison().contains('Leeds')
+
+    incidentDetailsPage.location().select('Asso A Wing')
+
+    incidentDetailsPage.cancelButton().should('not.exist')
+
+    incidentDetailsPage.clickSave()
+
+    checkAnswersPage.prison().contains('Leeds')
```
andrewrlee, line 98: This works but it would be better to verify that you are on the page.
 
We tend to use  the pattern:

```js
checkAnswersPage = CheckAnswersPage.verifyOnPage()
checkAnswersPage.prison().contains('Leeds')
```

integration-tests/integration/reportPages/change-prison.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    incidentDetailsPage.clickChangePrison()
+
+    const changePrisonPage = ChangePrisonPage.verifyOnPage()
+
+    changePrisonPage.changePrisonToLeeds()
+
+    changePrisonPage.clickCancel()
+
+    incidentDetailsPage.prison().contains('Moorland')
```
andrewrlee, line 116: and here - if you move the const to a let you can reuse `incidentDetailsPage `

integration-tests/integration/reportPages/enter-incident-details.spec.js
```diff
@@ -243,4 +244,12 @@ context('Submitting details page form', () => {
       .name()
       .should('not.exist')
   })
+  it('Displays the current prison', () => {
+    cy.login(bookingId)
+    fillFormAndSave()
+    cy.go('back')
```
andrewrlee, line 250: would avoid adding a specific test to this (ideally keep build times down) - could add this to `fillFormAndSave` when we check the offender names is set

integration-tests/mockApis/elite2api.js
```diff
@@ -102,6 +102,34 @@ module.exports = {
...
     })
   },
+  stubPrison: prisonId => {
+    const getPrisonDescription = () => {
+      const prisons = [
+        { id: 'LEI', description: 'Leeds' },
+        { id: 'MDI', description: 'Moorland' },
+      ]
+      return prisons.find(prison => prison.id === prisonId).description
+    }
```
andrewrlee, line 112: interesting, could remove the find and the method call by just defining descriptions as data structure: 
```js
const descriptions = {
  LEI: 'Leeds',
  MDI: 'Moorland'
}
...
 jsonBody: {
          agencyId: prisonId,
          description: description[agencyId],
          agencyType: 'INST',
          active: true,
        },
```

server/routes/changePrison.test.js
```diff
@@ -0,0 +1,42 @@
...
+      })
+  })
+
+  test('should contain Brixton as one of the options in the select', () => {
+    return request(app)
+      .get(`/report/-19/change-prison`)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Brixton')
+      })
```
andrewrlee, line 40: could do with a test for `post`

server/routes/createReport.js
```diff
@@ -116,13 +137,18 @@ module.exports = function NewIncidentRoutes({ reportService, offenderService, in
...
     const data = {
       ...input,
       displayName,
       offenderNo,
       incidentDate: getIncidentDate(incidentDate, input && input.incidentDate),
       locations,
       involvedStaff,
+      prison,
+      editMode,
+      bookingId,
```
andrewrlee, line 151: I'm not sure why we would need to add edit mode and booking id here? I would have thought they already existed?

andrewrlee: if you look at render form - they are added there I think?

andrewrlee: as an aside - I definitely think that we should look at extracting out an edit incident details page soon.. trying to handle most of this stuff in a generic way is getting more and more complicated and incident details is the main culprit

server/views/formPages/incident/incidentDetails.html
```diff
@@ -98,6 +99,23 @@ <h1 class="govuk-heading-xl mainHeading">{{ pageTitle }}</h1>
       }}
     </div>
   </div>
+  <div class="govuk-!-margin-bottom-6">
+    {% if data.editMode %}
+    {% set formName = "edit-change-prison" %}
+    {% else %}
+    {% set formName = "change-prison" %}
+    {% endif %}
```
andrewrlee, line 107: I don't think this is used?

integration-tests/integration/reportPages/change-prison.spec.js
GurnankCheema: done

integration-tests/integration/reportPages/change-prison.spec.js
GurnankCheema: not used independently
Done

integration-tests/integration/reportPages/change-prison.spec.js
GurnankCheema: done

integration-tests/integration/reportPages/change-prison.spec.js
GurnankCheema: done

integration-tests/integration/reportPages/enter-incident-details.spec.js
GurnankCheema: done

integration-tests/mockApis/elite2api.js
GurnankCheema: yes, simpler ...
done

server/routes/createReport.js
GurnankCheema: removed

server/views/formPages/incident/incidentDetails.html
GurnankCheema: correct.
done

server/routes/changePrison.test.js
GurnankCheema: done

integration-tests/integration/reportPages/change-prison.spec.js
andrewrlee: this hasn't been done?

server/services/locationService.ts
```diff
@@ -0,0 +1,31 @@
+import { Prison } from '../types/uof'
+
+import logger = require('../../log.js')
+
+import incidentClient = require('../data/incidentClient')
```
andrewrlee, line 5: this is a dependency so should be injected rather than required.

server/services/locationService.test.ts
```diff
@@ -0,0 +1,79 @@
...
+const elite2ClientBuilder = jest.fn()
+import incidentClient = require('../data/incidentClient')
+const token = 'token'
+
+const elite2Client = {
+  getPrisons: jest.fn(),
+  getPrisonById: jest.fn(),
+}
+
+incidentClient.updateAgencyId = jest.fn()
```
andrewrlee, line 12: if you overwrite this method like this, you do so for all other tests.
The correct way of mocking requires is to use jest.mock but the standard pattern we use is to inject it as a dependency.

(Could argue that it is surprising that the update of prison is on here - If I was looking for it, I would expect it to be on the report service)

server/services/locationService.ts
GurnankCheema: done

server/services/locationService.test.ts
GurnankCheema: done

integration-tests/integration/reportPages/change-prison.spec.js
GurnankCheema: missed in error.
Now done.

PR #340, opened 2020-06-18
DCS-460 moving to node 12

andrewrlee:


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 executors:
   builder:
     docker:
-      - image: circleci/node:10-buster-browsers
+      - image: circleci/node:12-buster-browsers
```
petergphillips, line 9: only very minor point - can use the one in the orb instead e.g. https://github.com/ministryofjustice/health-kick/blob/master/.circleci/config.yml#L9-L10

andrewrlee: thanks, will do in a follow up

PR #346, opened 2020-06-19
DCS 440 tweaks to check statement before sending page

GurnankCheema:


server/views/pages/statement/check-your-statement.html
```diff
@@ -1,4 +1,5 @@
 {% extends "../../partials/layout.html" %} 
+{% from  "../../macros.njk" import submitLink %}
```
andrewrlee, line 2: guess this isn't needed

GurnankCheema: well spotted

integration-tests/integration/statements/not-submit-statement.spec.js
```diff
@@ -0,0 +1,49 @@
+const YourStatementsPage = require('../../pages/yourStatements/yourStatementsPage')
+const WriteYourStatementPage = require('../../pages/yourStatements/writeYourStatementPage')
+const { ReportStatus } = require('../../../server/config/types')
+const CheckYourStatementPage = require('../../pages/yourStatements/checkYourStatementPage')
+
+context('Submit statement', () => {
```
andrewrlee, line 6: think this wasn't updated.. i'd have been tempted to have just added it as a test to the submit statement spec as it's not really worthy of its own "use case"

server/views/pages/statement/check-your-statement.html
```diff
@@ -39,15 +40,10 @@ <h1 class="govuk-heading-xl mainHeading">{{ pageTitle }} </h1>
...
-          name: 'submit',
-          href: '/',
-          classes: 'govuk-button govuk-button--secondary govuk-!-margin-left-3',
-          attributes: { 'data-qa': 'cancel' }
-        })
-      }}
+
+      <br/>
+      
+      <a href="/your-statements" class="govuk-link govuk-link--no-visited-state govuk-!-font-size-19" data-qa="save-and-complete-later">Save and complete later</a>
```
andrewrlee, line 45: I think you can use `govuk-body`, i honestly don't know which is best

integration-tests/integration/statements/not-submit-statement.spec.js
GurnankCheema: I did think of putting it in the submit-statement for the same reason that i's such a small test. But then I decided against because this test isn't about submitting a statement.

server/views/pages/statement/check-your-statement.html
GurnankCheema: i think the div that wraps the whole section (StatementDetailMacro  line 3) would cover it?
Plus it's in the Div that wraps the whole page, line 14

PR #352, opened 2020-06-23
DCS-406 Splitting out incident and review routes

andrewrlee:


server/routes/incidents.test.js
```diff
@@ -94,68 +58,4 @@ describe('GET /your-report', () => {
         expect(res.text).toContain('Use of force report')
       })
   })
-
-  it('should fail to render page when no report', () => {
```
brightonsbox, line 98: Do you not want to keep this test in for `/your-report`? This is now only testing the happy path for this endpoint now isn't it?

server/routes/reviewer.test.js
```diff
@@ -0,0 +1,102 @@
...
+    reviewService.getReport.mockReturnValue({ id: 1, form: { incidentDetails: {} } })
+    return request(app)
+      .get('/1/view-report')
+      .expect(200)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Use of force report')
+        expect(res.text).not.toContain('Delete report')
+      })
+  })
```
brightonsbox, line 67: Is this missing the `should fail to render page when no report` test?

server/routes/incidents.test.js
andrewrlee: the logic has been pushed down to the service - so we don't explicitly need to test it at this level, its just an unexpected error

server/routes/reviewer.test.js
andrewrlee: the logic has been pushed down to the service - so we don't explicitly need to test it at this level, its just an unexpected error

PR #354, opened 2020-06-29
DCS-406 Adding filter to all incidents page

andrewrlee:
Also tidying up integration tests.

Looks like:
<kbd>
Screenshot 2020-06-30 at 00 51 06
</kbd>


integration-tests/db/db.js
```diff
@@ -78,8 +94,8 @@ const submitStatement = ({ userId, reportId }) =>
       jobStartYear: 2017,
       statement: 'Things happened',
     })
-    .then(() => {
-      statementService.submitStatement(userId, reportId)
+    .then(async () => {
+      await statementService.submitStatement(userId, reportId)
```
bell-pa, line 98: I don't understand this.  Why would you mix `.then` with `async` and `await`? In fact, why use call-backs at all when `async/await` is available? I'm sure there's a good reason, but I can't see it.  (Perhaps coffee will help).

andrewrlee: this is a little bit flailing in the dark and its a bit weird 

There's an intermittent failure here where we call `submitStatement` and then refresh the page. 
When the test fails the refresh fails to show that the submit statement has gone through.

I think what's happening is that 
```
.then(() => {
      statementService.submitStatement(userId, reportId)
})
```
wasn't waiting for submit to occur. 

The following (flatmap) should work but doesn't compile for some reason in cypress
```
.then(() => statementService.submitStatement(userId, reportId))
```
so by using async it will recognise that the function is returning a promise and perform the flatmap/blocking appropriately.

Still not entirely sure that this fixes the issue as it's intermittent, and quite a few things I tried didn't!

bell-pa: Does it matter what value the handler returns? Would returning the promise from `submitStatement` do the trick?
```javascript
.then(() => statementService.submitStatement(userId, reportId)) 
```

PR #356, opened 2020-07-06
DT-993: Rename elite2 / auth

petergphillips:


andrewrlee:
Thanks but got a PR out that changes the service name as well. Will add the auth update to that.

PR #373, opened 2020-07-23
Dcs 567 split all incidents into two tabs

GurnankCheema:
The new screen are like this

_completed incident page with zero results_
<bkr>
completed incident with zero results
</bkr>

_not completed with zero result_
<bkr>
not completed with zer results
</bkr>

_not completed incident 1 result_
<bkr>
not completed incident
</bkr>

_completed incidents - 1 result_
<bkr>
completed with 1 result
</bkr>

_overdue label displayed_
<bkr>
not completed with overdue label
</bkr>


integration-tests/pages/reviewer/notCompletedIncidentsPage.js
```diff
@@ -0,0 +1,61 @@
...
+      viewStatementsButton: () => todoCol(i, 4).find('a'),
+      overdue: () => cy.get('[data-qa=overdue]'),
+      reportId: () =>
+        todoCol(i, 4)
+          .find('a')
+          .invoke('attr', 'href')
+          .then(link => link.match(/\/(.*?)\/view-statements/)[1]),
+    }),
+
+    // getCompleteRow: i => ({
```
andrewrlee, line 41: need to get rid of all this commented code

server/routes/coordinator.js
```diff
@@ -57,6 +57,9 @@ module.exports = function Index({ reportService, involvedStaffService, reviewSer
 
     confirmDeleteReport: async (req, res) => {
       const { reportId } = req.params
+      const referringPage = req.header('Referer') || 'confirm-delete'
```
andrewrlee, line 60: I would generally avoid using referrer if you can get away with it as its only transient information - any refresh or redirection will remove it

server/routes/reviewer.ts
```diff
@@ -81,3 +67,23 @@ export = function CreateReviewRoutes({ offenderService, reportDetailBuilder, rev
     },
   }
 }
+
+function viewIncidents(page, tab, reviewService) {
```
andrewrlee, line 71: I wouldn't try to avoid the duplication by keeping the 2 handler separate - the pages are considerably different. 
Don't need to support search on incomplete reports at all which leads to a very simple handler function.

server/routes/reviewer.ts
```diff
@@ -81,3 +67,23 @@ export = function CreateReviewRoutes({ offenderService, reportDetailBuilder, rev
...
+    const { prisonerName, prisonNumber, reporter, dateFrom, dateTo } = req.query
+    const query = removeKeysWithEmptyValues({
+      prisonerName,
+      prisonNumber,
+      reporter,
+      dateFrom: dateFrom ? parseDate(dateFrom, 'D MMM YYYY') : null,
+      dateTo: dateTo ? parseDate(dateTo, 'D MMM YYYY') : null,
+    })
+
+    const reports = await reviewService.getReports(res.locals.user.username, res.locals.user.activeCaseLoadId, query)
```
andrewrlee, line 82: Think we should be splitting this call out into 2 separate methods (getIncompleteReports and getCompletedReports) on the service - we don't want to return all incidents to the template. This will become even more apparent when trying to apply pagination

server/routes/coordinator.js
```diff
@@ -84,7 +87,7 @@ module.exports = function Index({ reportService, involvedStaffService, reviewSer
         await reportService.deleteReport(res.locals.user.username, reportId)
```
andrewrlee, line 87: so if you load the report before this line, (maybe line 85? as you will always need it whether confirmed or not)  you should be able to grab the status of the report and determine the correct location at this point only.

GurnankCheema: Yes, that could work

integration-tests/pages/reviewer/notCompletedIncidentsPage.js
GurnankCheema: yes, i will do a clean up. Just kept it there for reference

server/routes/reviewer.ts
GurnankCheema: ok

server/routes/coordinator.js
GurnankCheema: changed

GurnankCheema: done

server/routes/reviewer.ts
GurnankCheema: done

integration-tests/integration/coordinator/add-involved-staff.spec.js
```diff
@@ -82,10 +84,9 @@ context('A use of force coordinator can add involved staff', () => {
     viewStatementsPage = ViewStatementsPage.verifyOnPage()
     viewStatementsPage.return().click()
 
-    const allIncidentsPage = AllIncidentsPage.verifyOnPage()
-    allIncidentsPage.getTodoRows().should('have.length', 1)
-    allIncidentsPage.getNoCompleteRows().should('exist')
```
andrewrlee, line 87: no longer testing that the previous report is complete... this was the idea around adding a get report count DB method.

integration-tests/integration/coordinator/remove-involved-staff.spec.js
```diff
@@ -44,12 +45,12 @@ context('A use of force coordinator can remove involved staff', () => {
 
     seedReport()
 
-    let allIncidentsPage = AllIncidentsPage.goTo()
-    allIncidentsPage.getTodoRows().should('have.length', 1)
-    allIncidentsPage.getNoCompleteRows().should('exist')
+    let notCompletedIncidentsPage = NotCompletedIncidentsPage.goTo()
+    notCompletedIncidentsPage.getTodoRows().should('have.length', 1)
+    notCompletedIncidentsPage.getNoCompleteRows().should('not.exist')
```
andrewrlee, line 50: `getNoCompleteRows ` would never exist as "no completed rows" will never show on the incomplete page.

This function will need to be removed from the `notCompletedIncidentsPage` and can check using a new mechanism using the DB for accessing report count

integration-tests/integration/coordinator/remove-involved-staff.spec.js
```diff
@@ -85,17 +86,18 @@ context('A use of force coordinator can remove involved staff', () => {
     viewStatementsPage = ViewStatementsPage.verifyOnPage()
     viewStatementsPage.return().click()
 
-    allIncidentsPage = AllIncidentsPage.verifyOnPage()
-    allIncidentsPage.getNoTodoRows().should('exist')
-    allIncidentsPage.getCompleteRows().should('have.length', 1)
+    notCompletedIncidentsPage = NotCompletedIncidentsPage.verifyOnPage()
+    notCompletedIncidentsPage.getNoTodoRows().should('not.exist')
```
andrewrlee, line 90: this doesn't seem right - are you actually on the NotCompletedPage here? 
I think the "no incomplete rows" should show after making the report complete.

integration-tests/integration/reviewer/view-not-completed-incidents.spec.js
```diff
@@ -162,45 +143,17 @@ context('A use of force reviewer can view all incidents at the current agency',
       ],
     })
```
andrewrlee, line 144: can get rid of some of these report seeding - we only had 3 so we could do a bit of testing on search. As there isn't search any more its safe to get rid of them.

integration-tests/integration/reviewer/view-report.spec.js
```diff
@@ -58,14 +55,14 @@ context('view review page', () => {
       })
     )
 
-    const allIncidentsPage = AllIncidentsPage.goTo()
-    allIncidentsPage.getTodoRows().should('have.length', 2)
-    allIncidentsPage.getNoCompleteRows().should('exist')
+    const notCompletedIncidentsPage = NotCompletedIncidentsPage.goTo()
+    notCompletedIncidentsPage.getTodoRows().should('have.length', 2)
+    notCompletedIncidentsPage.getNoCompleteRows().should('not.exist')
```
andrewrlee, line 60: this doesn't make sense as notCompletedIncidentsPage would never show "No Complete Rows".

integration-tests/pages/reviewer/notCompletedIncidentsPage.js
```diff
@@ -21,7 +18,6 @@ const incidentsPage = () =>
     },
     getTodoRows: () => cy.get('[data-qa=incidents-todo]').find('tbody').find('tr'),
     getNoTodoRows: () => cy.get('[data-qa=no-incidents-todo]'),
-    getCompleteRows: () => cy.get('[data-qa=incidents-complete]').find('tbody').find('tr'),
     getNoCompleteRows: () => cy.get('[data-qa=no-incidents-complete]'),
```
andrewrlee, line 21: this should be deleted.

server/routes/coordinator.js
```diff
@@ -80,11 +79,14 @@ module.exports = function Index({ reportService, involvedStaffService, reviewSer
         return res.redirect(`/coordinator/report/${reportId}/confirm-delete`)
       }
 
+      const report = await reviewService.getReport(reportId)
+      const referringPage = report.status === 'SUBMITTED' ? '/not-completed-incidents' : '/completed-incidents'
```
andrewrlee, line 83: We have a constant you can refer to here: `ReportStatus.SUBMITTED.value`

`import { ReportStatus } from '../config/types'`

server/routes/reviewer.test.js
```diff
@@ -24,20 +25,23 @@ let app
...
+  reviewService.getIncompleteReports.mockResolvedValue({ awaiting: [] })
+  reviewService.getCompletedReports.mockResolvedValue([{}])
 })
 
 afterEach(() => {
   jest.resetAllMocks()
 })
 
-describe('GET /all-incidents', () => {
+const incidentStatus = ['/completed-incidents', '/not-completed-incidents']
```
andrewrlee, line 37: think this can be deleted

server/routes/reviewer.test.js
```diff
@@ -85,7 +89,29 @@ describe('GET /all-incidents', () => {
   it('should not allow if not reviewer', () => {
     userSupplier.mockReturnValue(user)
     return request(app)
-      .get('/all-incidents')
+      .get('/completed-incidents')
+      .expect(401)
+      .expect(res => expect(res.text).toContain('Not authorised to access this resource'))
```
andrewrlee, line 94: missing expectations on review service like line 82

server/routes/reviewer.test.js
```diff
@@ -85,7 +89,29 @@ describe('GET /all-incidents', () => {
...
+describe(`GET /not-completed-incidents`, () => {
+  it('should render page for reviewer', () => {
+    userSupplier.mockReturnValue(reviewerUser)
+
+    return request(app)
+      .get(`/not-completed-incidents`)
+      .expect(200)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Use of force incidents')
```
andrewrlee, line 107: missing expectations on review service like line 82

server/routes/reviewer.ts
```diff
@@ -54,8 +68,9 @@ export = function CreateReviewRoutes({ offenderService, reportDetailBuilder, rev
       )
 
       const statements = await reviewService.getStatements(await systemToken(res.locals.user.username), reportId)
+      const tab = report.status === 'COMPLETE' ? '/completed-incidents' : '/not-completed-incidents'
```
andrewrlee, line 71: Should use the ReportStatus constant as descibed in the other comment.

integration-tests/integration/coordinator/remove-involved-staff.spec.js
GurnankCheema: check for 'no completed rows' removed from spec.
getNoCompleteRows function removed from notCompletedIncidentsPage

integration-tests/integration/coordinator/remove-involved-staff.spec.js
GurnankCheema: Yes, you are correct. I  have moved this check to the end of the test.

integration-tests/integration/reviewer/view-not-completed-incidents.spec.js
GurnankCheema: Yes. There isn't a need for the 'filter' test because there is no search anymore. Test and seeding removed.

integration-tests/integration/reviewer/view-report.spec.js
GurnankCheema: agreed

integration-tests/pages/reviewer/notCompletedIncidentsPage.js
GurnankCheema: deleted

server/routes/coordinator.js
GurnankCheema: Constant used. 
Could not use the import statement because it was giving a babel error. 
Used _const { ReportStatus } = require('../config/types')_ instead

server/routes/reviewer.test.js
GurnankCheema: yes, unused variables

server/routes/reviewer.test.js
GurnankCheema: expectation added

server/routes/reviewer.ts
GurnankCheema: done

integration-tests/integration/coordinator/add-involved-staff.spec.js
GurnankCheema: testing the report is complete by navigating to the completed incidents page and checking it is empty

integration-tests/pages/reviewer/notCompletedIncidentsPage.js
```diff
@@ -38,24 +33,15 @@ const incidentsPage = () =>
...
-      reporter: () => completeCol(i, 3),
-      viewStatementsButton: () => completeCol(i, 4).find('a'),
-      reportId: () =>
-        completeCol(i, 4)
-          .find('a')
-          .invoke('attr', 'href')
-          .then(link => link.match(/\/(.*?)\/your-statement/)[1]),
-    }),
+    getCountOfNotCompleteReports: () => {
+      return cy.task('getCountOfNotCompleteReports', 'COMPLETE')
```
andrewrlee, line 37: tying this to a specific page really reduces the usability of this... tasks sit outside of the functionality of a page so would just use it directly in specs.

integration-tests/integration/coordinator/remove-involved-staff.spec.js
```diff
@@ -106,6 +105,10 @@ context('A use of force coordinator can remove involved staff', () => {
           { username: 'TEST_USER name', link: 'View statement', isOverdue: false, isUnverified: false },
         ])
       )
+
+    notCompletedIncidentsPage.getCountOfNotCompleteReports().then(count => {
+      expect(parseInt(count.rows[0].report_count, 10)).to.equal(0)
```
andrewrlee, line 110: would be good to push `parseInt(count.rows[0].report_count, 10))` into the task, as then you don't have to spread implementation details to everywhere that calls it.

integration-tests/pages/reviewer/notCompletedIncidentsPage.js
GurnankCheema: yes I noticed that approach had been used in integration-tests/integration/createIncident/form-submit.spec.js lines 70, 95

integration-tests/db/db.js
```diff
@@ -107,6 +107,13 @@ const clearAllTables = async () => {
   await deleteRows('report')
 }
 
+const getCountOfNotCompleteReports = reportStatus => {
+  return db.query({
+    text: `SELECT COUNT(r.status) as report_count FROM report r WHERE r.status != $1`,
+    values: [reportStatus],
+  })
```
andrewrlee, line 114: Ah... so you added it! 

`getCountOfNotCompleteReports` => `getReportCountWhereStatusIsNot` ? as the name is a bit inaccurate

might be better to flip it around: `getReportCount(status)`

`report_count` => `"reportCount"` (better to use javascript naming conventions as we'll be consuming it from java script)

And then also we have been using the report status enum, so passing in reportStatus enum and using the value here: `[reportStatus.value]`

integration-tests/integration/coordinator/add-involved-staff.spec.js
andrewrlee: now you have added the db method you can use it here?

integration-tests/db/db.js
GurnankCheema: i did try reportCount first but when the response arrived back to the spec, it was getting changed to reportcount, hence I used the uderscore

integration-tests/integration/coordinator/add-involved-staff.spec.js
GurnankCheema: DB method implemented to get a count of the reports that have a status of COMPLETE and asserted that to be zero as the test is to move the report from complete to incomplete

integration-tests/integration/coordinator/remove-involved-staff.spec.js
GurnankCheema: done

integration-tests/db/db.js
```diff
@@ -107,6 +107,15 @@ const clearAllTables = async () => {
   await deleteRows('report')
 }
 
+const getReportCount = reportStatus => {
+  return db
```
andrewrlee, line 111: ok so now the variable must take 2 parameters but the variable name is singular? 

We don't really want to hardcode the number of parameters and also the query is now doing something different when there is only 1 param passed - its says "it's either this value or null"

To get round this we use pg-format. 
`import format from 'pg-format'` (though const/require as JS)
 
then: 

```
.query({
    text: format(
      ` ...   from report r where r.status in (%L)`,
      statuses
    ),
```
where statuses is an array of status.

integration-tests/db/db.js
```diff
@@ -107,6 +107,15 @@ const clearAllTables = async () => {
...
+    .query({
+      text: `SELECT COUNT(r.status) as count FROM report r WHERE r.status IN ($1, $2)`,
+      values: [reportStatus[0], reportStatus[1]],
+    })
+    .then(result => parseInt(result.rows[0].count, 10))
```
andrewrlee, line 116: 👍

integration-tests/integration/coordinator/remove-involved-staff.spec.js
```diff
@@ -106,6 +105,10 @@ context('A use of force coordinator can remove involved staff', () => {
           { username: 'TEST_USER name', link: 'View statement', isOverdue: false, isUnverified: false },
         ])
       )
+
+    cy.task('getReportCount', [ReportStatus.SUBMITTED.value, ReportStatus.IN_PROGRESS.value]).then(count =>
```
andrewrlee, line 109: `seedReport` takes a ReportStatus and unwraps there rather than passing raw values through like this which is preferable but not the end of the world I guess.

integration-tests/db/db.js
GurnankCheema: yes, I was thinking of how to not hardcode the  array. 
Done

integration-tests/integration/coordinator/remove-involved-staff.spec.js
GurnankCheema: Yes. I did it this way to make the method generic so it could be used with different args. 
In `remove-involved-staff.spec.js` i've passed in SUBMITTED and IN_PROGRESS
but in `add-involved-staff.spec.js`, have passed in COMPLETE

PR #375, opened 2020-07-24
DCS-570 Merging completed and incomplete reports in 'Your reports'

andrewrlee:
<kbd>Screenshot 2020-07-24 at 15 13 43</kbd>


server/data/incidentClient.test.ts
```diff
@@ -33,34 +33,22 @@ describe('getCurrentDraftReport', () => {
...
-            , r.incident_date "incidentDate"
-            from v_report r
-          where r.status in ('IN_PROGRESS','SUBMITTED')
-          and r.user_id = 'user1'
-          order by r.incident_date desc`,
+          where r.user_id = $1
+          order by (case status 
+            when 'IN_PROGRESS' then 1
+            else 2
+            end), r.incident_date desc`,
```
andrewrlee, line 50: make sure IN_PROGRESS reports are first.. don't care about order of SUBMITTED or COMPLETED

server/data/incidentClientTypes.ts
```diff
@@ -13,6 +13,7 @@ export interface ReportSummary {
   reporterName: string
   offenderNo: string
   incidentDate: Date
+  status?: string
```
andrewrlee, line 16: this could be made mandatory after Gurnank finishes the bit of work he is doing

server/services/reportService.test.ts
```diff
@@ -1,28 +1,21 @@
...
-  getReportForReviewer: jest.fn(),
-  getReportsForReviewer: jest.fn(),
-  updateDraftReport: jest.fn(),
-  createDraftReport: jest.fn(),
-  submitReport: jest.fn(),
-  deleteReport: jest.fn(),
-  commitAndStartNewTransaction: jest.fn(),
-  getIncompleteReportsForReviewer: jest.fn(),
-  getCompletedReportsForReviewer: jest.fn(),
-  updateAgencyId: jest.fn(),
```
andrewrlee, line 18: nice replacing manual mocks with generated class mocks

server/services/reportService.test.ts
```diff
@@ -1,28 +1,21 @@
...
-  inTransaction: fn => fn(client),
-}
+const incidentClient = new IncidentClient(jest.fn as any, jest.fn() as any) as jest.Mocked<IncidentClient>
+
+const offenderService = {
+  getOffenderNames: jest.fn(),
+  getOffenderImage: jest.fn(),
+  getOffenderDetails: jest.fn(),
+  getPrisonersDetails: jest.fn(),
+} as jest.Mocked<OffenderService>
```
andrewrlee, line 15: can't generate a mock class for this as OffenderService is an interface. Do get compilation failure if it doesn't match the shape of the service though which is nice

server/services/reportService.test.ts
```diff
@@ -155,73 +149,83 @@ describe('getReport', () => {
...
     const formObject = { decision: 'Yes', followUp1: 'County', followUp2: 'Town' }
-    const involvedStaff = [{ name: 'Bob', username: 'BOB', email: 'bob@gov.uk' }]
 
     await service.update({
       currentUser,
       bookingId: 1,
       formId: 'form1',
       formObject,
       incidentDate: { value: null },
-      involvedStaff,
```
andrewrlee, line 208: now get compilation errors as this parameter never actually existed

server/services/reportService.ts
```diff
@@ -0,0 +1,188 @@
...
+    const token = await this.systemToken(username)
+    const offenderNos = incidents.map(incident => incident.offenderNo)
+    return this.offenderService.getOffenderNames(token, offenderNos)
+  }
+
+  getCurrentDraft(userId, bookingId) {
+    return this.incidentClient.getCurrentDraftReport(userId, bookingId)
+  }
+
+  getReportStatus(report) {
```
brightonsbox, line 59: Why does this need wrapping like this?

andrewrlee: the function is imported from another module. mixing in this functionality as it feels like it belongs here and its convenient to mock out the entire service when included as a dependency rather than 2 separate parts.

PR #382, opened 2020-08-03
DCS-572 fix issue where search parameters were not being passed through to subsequent pages

andrewrlee:


server/routes/reviewer.ts
```diff
@@ -48,6 +48,7 @@ export = function CreateReviewRoutes({ offenderService, reportDetailBuilder, rev
         reports,
         pageData,
         query,
```
andrewrlee, line 50: couldn't use this query as it has `moment`s in which gets converted to a blank value in the query string

PR #385, opened 2020-08-26
DCS-562 Hook up token verification service

bell-pa:


server/services/healthcheck.js
```diff
@@ -13,8 +13,13 @@ const service = (name, url) => {
       .catch(err => ({ name, status: 'ERROR', message: err }))
 }
 
-module.exports = function healthcheckFactory(authUrl, elite2Url) {
-  const checks = [db, service('auth', `${authUrl}/ping`), service('elite2', `${elite2Url}/ping`)]
+module.exports = function healthcheckFactory(authUrl, elite2Url, tokenVerificationUrl) {
+  const checks = [
+    db,
+    service('auth', `${authUrl}/ping`),
+    service('elite2', `${elite2Url}/ping`),
```
andrewrlee, line 20: guess these should probably call /heath/ping too

andrewrlee (approved):
LGTM

server/authentication/auth.js
```diff
@@ -1,16 +1,18 @@
...
+    if (req.isAuthenticated() && (await tokenVerifier.verify(res.locals.user.token))) {
       return next()
     }
-
-    req.session.returnTo = req.originalUrl
-    return res.redirect('/login')
+    req.logout()
+    const query = querystring.stringify({ returnTo: req.originalUrl })
+    res.redirect(`/login?${query}`)
+    return undefined // because typescript insists on a return value
```
andrewrlee, line 15: kind of standard to just `return res.redirect(`/login?${query}`)` in express apps

server/authentication/auth.js
```diff
@@ -1,16 +1,18 @@
...
-function authenticationMiddleware() {
-  return (req, res, next) => {
-    if (req.isAuthenticated()) {
+function authenticationMiddlewareFactory(tokenVerifier) {
+  return async (req, res, next) => {
+    if (req.isAuthenticated() && (await tokenVerifier.verify(res.locals.user.token))) {
       return next()
     }
-
-    req.session.returnTo = req.originalUrl
```
andrewrlee, line 12: this is used by `/login/callback` in app.js. Do we need to remove/change that?

docker-compose.yml
```diff
@@ -11,7 +11,46 @@ services:
...
+      SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI: "http://auth-server:8080/auth/.well-known/jwks.json"
+
+  token-verification-api-db:
+    image: redis
+    networks:
+      - token-backend
+    container_name: token-verification-api-db
+    restart: always
+    ports:
+      - '6379:6379'
```
andrewrlee, line 43: both these images sit on the same port? 

Would it be possible to use the same redis instance for both use cases?

andrewrlee:
Think you might be missing this: https://github.com/ministryofjustice/licences/pull/1226/files

docker-compose.yml
bell-pa: Separate networks. But, yes,  we could share the Redis instance.  I don't _think_ it would change any behaviour...

server/authentication/auth.js
bell-pa: No

server/authentication/auth.js
bell-pa: Done

server/services/healthcheck.js
bell-pa: Done

andrewrlee (approved):

PR #386, opened 2020-08-26
Dcs 583 make date and time of incident a mandatory field

andrewrlee:
Also updating cypress and a couple of other libraries


server/views/formPages/timeMacro.njk
```diff
@@ -0,0 +1,85 @@
+{% from "govuk/components/error-message/macro.njk" import govukErrorMessage %}
```
andrewrlee, line 1: we'll replace this is a follow up with something simpler

server/config/forms/incidentDetailsForm.js
```diff
@@ -2,44 +2,54 @@ const { joi, asMeta, validations, usernamePattern, namePattern, caseInsensitiveC
...
+      minute: joi
+        .any()
+        .custom(minuteValidation)
+        .messages({
+          [ValidationError.missing]: 'Enter missing minutes',
+          [ValidationError.isNotNumber]: 'Enter minutes using numbers only',
+          [ValidationError.isTooLarge]: 'Enter the minutes using 59 or less',
+          [ValidationError.isNot2Digits]: 'Enter the minutes using 2 digits',
+        }),
+    }).custom(timeValidation).messages({
```
andrewrlee, line 48: Validation on the "time" field isn't directly mapped to any specific input, instead it just targets the overall time component which is a bit of a compromise as if we target a specific input, we can't highlight additional elements:
In the case of time in the future ideally we'd highlight both hours and minutes and clicking the link would take you to the hour input.

assets/sass/local.sass
```diff
@@ -292,6 +292,9 @@ dt.summary-list__key__wider
     width: 150px
     margin-bottom: 32px
 
+.date-picker-container span
+    width: 850px
```
Nimphal, line 296: Out of curiosity, what's the reason for this?

integration-tests/integration/reportPages/enter-incident-details.spec.js
```diff
@@ -19,6 +19,10 @@ context('Submitting details page form', () => {
   const fillFormAndSave = () => {
     const reportUseOfForcePage = ReportUseOfForcePage.visit(offender.bookingId)
     const incidentDetailsPage = reportUseOfForcePage.startNewForm()
+
+    incidentDetailsPage.incidentDate.date().type('12/01/2020{esc}')
```
Nimphal, line 23: You're joking... Had no idea you could just include esc in the input and it will close it!

package.json
```diff
@@ -138,12 +139,13 @@
...
   "devDependencies": {
     "@types/express": "^4.17.7",
+    "@types/hapi__joi": "^17.1.4",
     "@types/jest": "^26.0.5",
     "@types/pg": "7.14.4",
     "@typescript-eslint/eslint-plugin": "^3.7.0",
     "@typescript-eslint/parser": "^3.7.0",
     "concurrently": "^5.2.0",
-    "cypress": "^4.12.0",
+    "cypress": "^5.0.0",
```
Nimphal, line 148: ooooh, shiny!

Nimphal (approved):
LGTM

assets/sass/local.sass
andrewrlee: I am not sure, a lot of this wasn't done by me. I'll have a look at this in a follow up..

PR #387, opened 2020-09-01
Dcs 583 make date and time of incident a mandatory field

GurnankCheema:
User no longer able to enter negative values for hours and minutes.
Future date now includes 23:59 of today


GurnankCheema:
new branch created

PR #388, opened 2020-09-01
DCS-583: User was previously able to input 23:59 minutes in current d…

GurnankCheema:
…ay without it failing future time validation.

This was because the timeValidation() function was working with minutes rather than seconds.
Function changed to use seconds.
The user could also previously enter negative numbers for hour and minutes. Validation added in minuteValidation() and hourValidation() to identify values less than zero.
dateSanitiser stoDateTime() updated to invalidate negative numbers

### DIFF
--- a/server/config/forms/incidentDateValidation.ts
+++ b/server/config/forms/incidentDateValidation.ts
@@ -9,6 +9,7 @@ export enum ValidationError {
   isNot2Digits = 'isNot2Digits',
   isFuture = 'isFuture',
   invalid = 'invalid',
+  isNotPositiveNumber = 'isNotPositiveNumber',
 }
 
 export function dateValidation(date: string, helpers: joi.CustomHelpers) {
@@ -29,7 +30,7 @@ export function dateValidation(date: string, helpers: joi.CustomHelpers) {
 export function timeValidation(time: string, helpers: joi.CustomHelpers) {
   const fullDate = helpers.state.ancestors[0].value
 
-  if (fullDate && moment(fullDate).isBetween(moment.now(), moment().endOf('day'), 'minutes')) {
+  if (fullDate && moment(fullDate).isBetween(moment.now(), moment().endOf('day'), 'seconds')) {
     return helpers.error(ValidationError.isFuture)
   }
 
@@ -45,6 +46,10 @@ export function minuteValidation(minute: string, helpers: joi.CustomHelpers) {
 
   if (parsedMinutes === null) {
     return helpers.error(ValidationError.isNotNumber)
+  }
+
+  if (parsedMinutes < 0) {
+    return helpers.error(ValidationError.isNotPositiveNumber)
   }
 
   if (parsedMinutes >= 60) {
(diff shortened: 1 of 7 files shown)


server/utils/dateSanitiser.test.js
```diff
@@ -100,6 +100,34 @@ describe('sanitiser', () => {
...
+
+    test('hour is less then 0', () =>
+      check({ date: '21/01/2019', time: { hour: '-1', minute: '0' } }).toEqual({
+        date: '21/01/2019',
+        time: { hour: '-1', minute: '0' },
+        value: null,
+      }))
+
+    test('minutes greater than 24', () =>
+      check({ date: '21/01/2019', time: { hour: '24', minute: '0' } }).toEqual({
```
andrewrlee, line 119: minutes are not great than 24?

server/config/forms/incidentDetailsForm.js
```diff
@@ -41,6 +41,7 @@ const requiredIncidentDate = joi
             [ValidationError.isNotNumber]: 'Enter hours using numbers only',
             [ValidationError.isTooLarge]: 'Enter an hour which is 23 or less',
             [ValidationError.isNot2Digits]: 'Enter the hours using 2 digits',
+            [ValidationError.isNotPositiveNumber]: 'Enter the hours using positive values only',
```
andrewrlee, line 44: `Enter an hour which is 0 or more` might be more in keeping with the tooLarge message?

did you check with content about this?

server/config/forms/incidentDateValidation.ts
```diff
@@ -29,7 +30,7 @@ export function dateValidation(date: string, helpers: joi.CustomHelpers) {
 export function timeValidation(time: string, helpers: joi.CustomHelpers) {
   const fullDate = helpers.state.ancestors[0].value
 
-  if (fullDate && moment(fullDate).isBetween(moment.now(), moment().endOf('day'), 'minutes')) {
+  if (fullDate && moment(fullDate).isBetween(moment.now(), moment().endOf('day'), 'seconds')) {
```
andrewrlee, line 33: Is this needed? I don't see a test that corresponds to this change and why it would be necessary?

GurnankCheema: So, minutes was allowing the user to input 23.59 form the current day and i was being accepted. Using seconds prevents that

server/utils/dateSanitiser.test.js
GurnankCheema: meant to be Greater than 59

server/config/forms/incidentDetailsForm.js
GurnankCheema: I didn't but I don't think there is a need to. I think the change is small enough to implement without them

GurnankCheema: I didn't but I don't think there is a need to. I think the change is small enough to implement without them

server/config/forms/incidentDetailsValidation.test.js
```diff
@@ -666,6 +793,40 @@ describe("'complete' validation", () => {
...
+          time: { hour: endOfToday.format('HH'), minute: endOfToday.format('mm') },
+        },
+      })
+      expect(formResponse).toEqual({
+        locationId: -1,
+        plannedUseOfForce: true,
+        involvedStaff: [{ username: 'ITAG_USER' }],
+        witnesses: [{ name: 'User bob' }],
+      })
+    })
```
andrewrlee, line 829: nice

andrewrlee (approved):
LGTM

PR #392, opened 2020-09-04
DCS-618: timeInput macro refactored

GurnankCheema:
The macro has been trimmed down to remove any unused elements. The incidentsDetails.html file uses this macro so that has been updated to pass in only the dynamic content (value and error classes). All the other classes and attributes for the hour and minute has been hard coded into the macro.
Have added a new class to widen the date input field because when there is an error in the date, the red errors border covers some of the field content.
Screen shot of errors below
<bkr>
dcs 618
</bkr>

### DIFF
--- a/server/views/formPages/incident/incidentDetails.html
+++ b/server/views/formPages/incident/incidentDetails.html
@@ -15,7 +15,7 @@
 {% block formItems %}
 <div class="govuk-body">
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds"> 
+    <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-xl mainHeading">{{ pageTitle }}</h1>
       <div class="govuk-!-margin-bottom-6">
         <p class="govuk-!-margin-bottom-1">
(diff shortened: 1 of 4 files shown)


server/views/formPages/timeMacro.njk
```diff
@@ -6,80 +6,61 @@
...
+            legend: {
+                text: "Time",
+                isPageHeading: false
+            }
+        }) 
+    %}
+
+    {{
+    govukHint({
+        id: 'incidentDate[time]-hint',
```
andrewrlee, line 21: I think we need to change the name of the file/macro - the names suggest a generic time macro but we've now hardcoded it to only refer to incident date time

server/views/formPages/incident/incidentDetails.html
```diff
@@ -24,54 +24,34 @@ <h1 class="govuk-heading-xl mainHeading">{{ pageTitle }}</h1>
...
-            legend: {
-              text: "Time",
-              isPageHeading: false
-            }
-          }, 
           errorMessage: errors | concatErrors(["incidentDate[time][hour]", "incidentDate[time][minute]", "incidentDate[time]"]),
-          items: [
-              {
-              label: 'Hours',
+          hour:  {
```
andrewrlee, line 47: nice, hardcoding names of the fields.

server/views/formPages/timeMacro.njk
andrewrlee: I might have been tempted to keep it generic and pass in an id, and name, value props for hours/minutes..

GurnankCheema: 1. file and macro name changed
2. id and name values now passed in to macro rather than being hard coded

andrewrlee (approved):
LGTM

PR #397, opened 2020-09-10
DCS 614: user can print the statement they submitted

GurnankCheema:
/your-statements -> /{reportId}/your-statement has the print link

**Page to print**

Screenshot 2020-09-10 at 16 05 52
</img>

**the formatted contents to print**

Screenshot 2020-09-10 at 16 06 02
</img>


server/data/statementsClientTypes.ts
```diff
@@ -8,3 +8,18 @@ type StatementSummary = {
   isOverdue: boolean
   status: string
 }
+
+type Statement = {
+  id?: number
```
andrewrlee, line 13: I think some of these don't need to be optional?
including : id, bookingId, name, reporterName, incidentDate

server/data/statementsClientTypes.ts
```diff
@@ -8,3 +8,18 @@ type StatementSummary = {
...
+  lastTrainingMonth?: number
+  lastTrainingYear?: number
+  jobStartYear?: number
+  statement?: string
+  submittedDate?: Date
+  name?: string
+  reporterName?: string
+}
+
+type AdditionalComments = Array<{ additionalComment: string; dateSubmitted: Date }>
```
andrewrlee, line 25: Good use of generics but it might be better to create a type for the a singular AdditionalComment as it might be more reusable. 

rather than `Array<AdditionalComment>` you can also just do `AdditionalComment[]`

server/services/statementServiceTypes.ts
```diff
@@ -0,0 +1,18 @@
...
+  lastTrainingMonth?: number
+  lastTrainingYear?: number
+  jobStartYear?: number
+  statement?: string
+  submittedDate?: Date
+  name?: string
+  reporterName?: string
+}
+
+type Status = {
```
andrewrlee, line 15: as I said there is a real type that could be surfaced. We need to change the types in `server/config/types.js` to enums and then we could explicitly refer to `StatementStatus` as a type in its own right. 

This is fine for now though and better than leaving untyped! I often have made mistakes with these fake javascript enums

server/services/statementServiceTypes.ts
```diff
@@ -0,0 +1,18 @@
+type UserStatement = {
+  additionalComments?: Array<{ additionalComment: string; dateSubmitted: Date }>
+  id?: number
```
andrewrlee, line 3: The same thing applies regarding optionality here.. 

but one thing we could do is use the types defined in statementClientTypes to define a new type:

So the main goal is "we want to express a type defined as a statement but also with comments".

For this, rather than duplicating the existing structure we could use [intersection types](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html).
```ts
import type { AdditionalComment, Statement } from '../data/statementsClientTypes'

type UserStatement = Statement & { additionalComments: AdditionalComment[] }

type Status = {
  value: string
  label: string
}
```

Might need to export the `statementClientType` members

And maybe `StatementWithComments` might be a better name?

assets/sass/local.sass
```diff
@@ -344,3 +344,31 @@ dt.summary-list__key__wider
...
+    top: 10px
+    padding: 0.5em 0.5em 0.5em 38px
+    background: url(/assets/images/Printer_icon.png) no-repeat 5px 50%
+    background-size: 22px 24px
+
+.print
+  display: none !important
+
+@media print
+  .no-print, .govuk-header, .govuk-footer
```
andrewrlee, line 365: is this going to apply to every page and so hide the header and footers from all print screens?
If so, Is it a problem?

assets/sass/local.sass
```diff
@@ -344,3 +344,31 @@ dt.summary-list__key__wider
...
 .text-align-center
   text-align: center !important
+
+.margin-top-bottom-0
+  margin-top: 0
+  margin-bottom: 0
+
+.print-link 
+    display: inline-block
+    margin: 0 0 15px -10px
```
andrewrlee, line 354: Think. it might generally be good practice to use [gov uk spacings](https://design-system.service.gov.uk/styles/spacing/) rather than hard coding pixel sizes:

GurnankCheema: Good idea. 
done.

assets/sass/local.sass
GurnankCheema: Yes, it will apply to any printed pages. I did consider whether there would be a reason where we would show the on-screen header and footer when printing something but I don't think we will need to. But I'll check with designers whether printing the header/footers will ever be needed

server/data/statementsClientTypes.ts
GurnankCheema: agreed. done

server/services/statementServiceTypes.ts
GurnankCheema: ok

server/data/statementsClientTypes.ts
GurnankCheema: done

andrewrlee (approved):
LGTM

PR #400, opened 2020-09-14
DCS-403 Removing missing user check

andrewrlee:
Previously we would prevent a user from submitting a report if they
had listed users in the report that didn't exist in nomis.

This was to handle an edge case where the user failed to resolve
a missing user after being prompted.

Going forward there will be no way to add a missing staff member so
this check will not be necessary. Implementing this change now
to ease moving to new flow. 

In the meanwhile we will automatically remove missing users for the user
(which should be fine as the user will have had to navigate away after being presented a warning message)


GurnankCheema (approved):
LGTM

PR #401, opened 2020-09-15
DCS-403 extracting out draft client

andrewrlee:
The idea being that operations that happen during a report being built and submitted are handled by this new client, whereas operations afterwards get handled by the incident client

### DIFF
--- a/server/data/draftReportClient.ts
+++ b/server/data/draftReportClient.ts
@@ -1,0 +1,98 @@
+import { QueryPerformer, InTransaction } from './dataAccess/db'
+import { ReportStatus } from '../config/types'
+import { DraftReport, NoDraftReport, StaffDetails } from './draftReportClientTypes'
+import { AgencyId } from '../types/uof'
+
+const maxSequenceForBooking =
+  '(select max(r2.sequence_no) from report r2 where r2.booking_id = r.booking_id and user_id = r.user_id)'
+
+export default class DraftReportClient {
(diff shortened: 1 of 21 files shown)


server/data/draftReportClient.ts
```diff
@@ -0,0 +1,98 @@
...
+import { AgencyId } from '../types/uof'
+
+const maxSequenceForBooking =
+  '(select max(r2.sequence_no) from report r2 where r2.booking_id = r.booking_id and user_id = r.user_id)'
+
+export default class DraftReportClient {
+  constructor(private readonly query: QueryPerformer, private readonly inTransaction: InTransaction) {}
+
+  async create({ userId, bookingId, agencyId, reporterName, offenderNo, incidentDate, formResponse }): Promise<number> {
+    const nextSequence = `(select COALESCE(MAX(sequence_no), 0) + 1 from v_report where booking_id = $5 and user_id = $2)`
```
bell-pa, line 13: Will this query work if the most recent report has been deleted ?  It looks as if it might return the sequence number of the deleted report.

andrewrlee: I might be missing something, it's going by the `v_report` view?

andrewrlee: not saying that there definitely isn't an issue, but this was just moved across from the other incident client

bell-pa (approved):
LGTM, but see comment on draftReportClient.ts line 13. Will this fail if the most recent report has been deleted?  Perhaps fix by selecting from report instead of v_report?

PR #403, opened 2020-09-16
Alee/dcs 403 refactor update draft

andrewrlee:


package.json
```diff
@@ -75,7 +75,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged && tsc && npm test"
```
andrewrlee, line 78: removing tests as takes >25seconds

server/routes/createReport.ts
```diff
@@ -66,30 +66,15 @@ export default class CreateReport {
...
-      })
-
-      if (updatedPayload) {
-        await this.draftReportService.update({
-          currentUser: res.locals.user,
-          formId,
-          bookingId: parseInt(bookingId, 10),
-          formObject: updatedPayload || {},
-        })
-      }
```
andrewrlee, line 92: pushing this logic down (and also duplicated logic in incident details) into service layer

server/services/report/draftReportService.ts
```diff
@@ -27,8 +27,14 @@ export default class DraftReportService {
...
-  public update(params: UpdateParams): Promise<number> {
-    return this.updateDraftReport.update(params)
+  public process(
+    currentUser: LoggedInUser,
+    bookingId: number,
+    formName: string,
+    updatedSection: unknown,
+    incidentDate?: Date | null
+  ): Promise<void> {
+    return this.updateDraftReport.process(currentUser, bookingId, formName, updatedSection, incidentDate)
```
andrewrlee, line 37: this just delegates down - not sure if there is a way of defining the args as a type so don't have to duplicate, something like: 


public process(
    currentUser: LoggedInUser,
    bookingId: number,
    formName: string,
    updatedSection: unknown,
    incidentDate?: Date | null
  ): Promise<void> {
    return this.updateDraftReport.process(currentUser, bookingId, formName, updatedSection, incidentDate)
 

 
this just delegates down - not sure if there is a way of defining the args as a type so don't have to duplicate, something like a typesafe version of: 
```ts
  public process(
    ...args: UpdateParams
  ): Promise<void> {
    return this.updateDraftReport.process(...args)
}
```

server/services/validation/formProcessing.js
```diff
@@ -9,25 +8,6 @@ const processInput = ({ validationSpec, input }) => {
   return { payloadFields, extractedFields, errors }
 }
 
-/**
- * Create new object from formObject with field 'formName' replaced by formPayload.
```
andrewrlee, line 13: this is now called `getUpdatedReport` and in the draft service

server/types/express.d.ts
```diff
@@ -5,17 +5,27 @@ declare namespace Express {
       token: string
       refreshToken: string
       refreshTime: any
-      firstName: string
-      lastName: string
-      userId: string
-      displayName: string
-      isReviewer: boolean
-      isCoordinator: boolean
-      activeCaseLoadId: string
```
andrewrlee, line 14: all this was on `res.locals.user`. There is a cut down version of the user on `req.user`

server/services/report/draftReportService.ts
andrewrlee: might be able to do this in [typescript 3 actually](https://auth0.com/blog/typescript-3-exploring-tuples-the-unknown-type/#:~:text=A%20tuple%20is%20a%20TypeScript,need%20not%20be%20the%20same).

bell-pa (approved):
LGTM

PR #410, opened 2020-09-25
Alee/dcs 640

andrewrlee:


bell-pa (approved):
I checked out the branch and tried running various scenarios I couldn't break anything. 
The Yes/No radio buttons in the'Staff involved in use of force' page seem a little odd. It might work better as a link / button 'Add another member of staff'.

Anyway running the app helped with the review.

LGTM 
🤞

PR #413, opened 2020-09-29
DCS-640 Remove content security policy. 

andrewrlee:
This is a middleware that is added by default version 4.0 of helmet. This prevents running of js unless it's specifically accounted for as part of the security policy which breaks everything. 

Added a ticket to the backlog to get it working under CSP


petergphillips (approved):
LGTM

PR #416, opened 2020-09-30
DCS-630 fix navigation around editing involved staff

andrewrlee:
Also extracting inline javascript for print links out


bell-pa (approved):
LGTM

PR #421, opened 2020-10-09
DCS-410 prevented coordinators from being able to remove the reporter…

louisenorris:
…s as a member of involved staff from the report and made sure reporter appears first


bell-pa (approved):
LGTM

PR #423, opened 2020-10-15
DCS-647 Add staff through user search and support multiple matches.

bell-pa:
Fixing problems found in peer test.


louisenorris (approved):
LGTM

PR #424, opened 2020-10-16
DCS-647 implemented content security policy for script tags

louisenorris:


assets/js/tag-manager.js
```diff
@@ -0,0 +1,15 @@
+dataLayer = [{ activeCaseloadId: '{{ user.activeCaseLoadId }}', userId: '{{ user.username | MD5 | upper }}' }]
```
brightonsbox, line 1: We might have a problem with this actually, I hadn't spotted that it has nunjucks variables passed to it.

One way to resolve this is to have variables in the html like so:

```
<input type="hidden" id="active-caseload" value='{{ user.activeCaseLoadId }}'>
<input type="hidden" id="user-id" value='{{ user.username | MD5 | upper }}'>
```

And then in this javascript file get the variables like so:

```
dataLayer = [{ activeCaseloadId: $('#active-caseload').val(), userId: $('#user-id').val() }]
```

assets/js/tag-manager.js
```diff
@@ -0,0 +1,15 @@
...
+//Google Tag Manager
+;(function (w, d, s, l, i) {
+  w[l] = w[l] || []
+  w[l].push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' })
+  var f = d.getElementsByTagName(s)[0],
+    j = d.createElement(s),
+    dl = l != 'dataLayer' ? '&l=' + l : ''
+  j.async = true
+  j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl + '{{ googleTagManagerEnvironment | safe }}'
+  f.parentNode.insertBefore(j, f)
```
bell-pa, line 12: Similar problem here and on next line U tgubj,.

assets/js/tag-manager.js
bell-pa: Would jquery need to load first for this to work?
Or... add these values to the script server-side with a cache bust in the markup.  At the page level? Might be OK at session level.

bell-pa (changes requested):
See comments on assets/js/tag-manager.js

assets/js/tag-manager.js
brightonsbox: Ah you're right, looks like this gets loaded before jquery, I suppose it can be done without though,
`document.getElementById("active-caseload").value` etc.

louisenorris:
Duplicate branch.

PR #428, opened 2020-10-26
DCS-658 Removing logging of stack traces in production

andrewrlee:
This was incorrectly logging stack traces in production.

Also removed request logging of assets. (These are still logged as `requests` by the standard instrumentation that bunyan)


server/app.ts
```diff
@@ -254,10 +254,10 @@ export default function createApp(services: Services): Express {
   app.use('/api/', createApiRouter(authenticationMiddleware, services))
 
   app.use((req, res, next) => {
-    next(new Error('Not found'))
+    next(createError(404, 'Not found'))
```
andrewrlee, line 257: providing correct error status

server/app.ts
```diff
@@ -149,6 +147,8 @@ export default function createApp(services: Services): Express {
     app.use('/assets/images/icons', express.static(path.join(process.cwd(), dir), cacheControl))
   })
 
+  app.use(bunyanRequestLogger({ name: 'Use of force http', serializers: loggingSerialiser }).requestLogger())
```
andrewrlee, line 150: moving this down so we don't log every request for assets as `trace`, will still do as `requests` though

bell-pa (approved):
LGTM

PR #429, opened 2020-10-26
DCS-663 correcting link

andrewrlee:
These links all go to the "Report use of force" page so do not need the incidents suffix


andymarke (approved):
LGTM

PR #431, opened 2020-10-27
DCS-664 Fixing sending reminders

andrewrlee:
Refactoring lead to wrong import being used. Fixed and moved area to typescript (which would have caught the issue)


job/sendReminders.ts
```diff
@@ -0,0 +1,41 @@
...
+/*
+ * Do appinsights first as it does some magic instrumentation work, i.e. it affects other 'require's
+ * In particular, applicationinsights automatically collects bunyan logs
+ */
+import { initialiseAppInsights, buildAppInsightsClient } from '../server/utils/azure-appinsights'
+
+initialiseAppInsights()
+
+import StatementsClient from '../server/data/statementsClient'
+import IncidentClient from '../server/data/incidentClient'
```
andrewrlee, line 11: the actual bug was here: was pulling in draftReportClient instead

server/utils/nunjucksSetup.ts
```diff
@@ -16,7 +17,7 @@ type Error = {
   text: string
 }
 
-export default function (app, path) {
+export default function configureNunjucks(app: Express.Application): void {
```
andrewrlee, line 20: fixing some type issues here and noticed path didn't need to be injected as a dependency

JamieDraperUK (approved):
LGTM

bell-pa (approved):
LGTM

PR #437, opened 2020-11-25
DCS-722 prevent content shift following 'add another'

GurnankCheema:
Also did tidy up of html missed in another ticket


andrewrlee (approved):
LGTM

PR #438, opened 2020-11-27
DCS-723 - reduced space between page title, body text and button by half

GurnankCheema:
screenshot

**before**
723 delete before

**after**
723 delete after


andrewrlee (approved):
LGTM

PR #440, opened 2020-11-30
DCS-727 CSS changes to change width of date fields.

GurnankCheema:
The css classes that have been removed are not used elsewhere in the app so this change does not affect any other pages

screenshots
**before**

727-before

727 S5 before

**after**

727-after

727 S5 after


andrewrlee (approved):
LGTM

PR #441, opened 2020-12-01
LNorris/DCS-725 added conditional to write-your-statement html to con…

louisenorris:
…trol error text alignment for refresher training month and year inputs


GurnankCheema (approved):
LGTM

PR #446, opened 2021-02-01
DCS-812: login integration test added. 

GurnankCheema:
Test needed to verify the application logs current user out if a new user logs in.

### DIFF
--- a/integration-tests/plugins/index.js
+++ b/integration-tests/plugins/index.js
@@ -13,7 +13,8 @@ module.exports = on => {
 
     getLoginUrl: auth.getLoginUrl,
 
-    stubLogin: () => Promise.all([auth.stubLogin({}), elite2api.stubUser(), elite2api.stubUserCaseloads()]),
+    stubLogin: (user = {}) =>
+      Promise.all([auth.stubLogin(user), elite2api.stubUser(user), elite2api.stubUserCaseloads()]),
 
     stubReviewerLogin: () =>
       Promise.all([auth.stubLogin({ isReviewer: true }), elite2api.stubUser(), elite2api.stubUserCaseloads()]),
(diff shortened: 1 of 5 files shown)


integration-tests/plugins/index.js
```diff
@@ -13,7 +13,7 @@ module.exports = on => {
 
     getLoginUrl: auth.getLoginUrl,
 
-    stubLogin: () => Promise.all([auth.stubLogin({}), elite2api.stubUser(), elite2api.stubUserCaseloads()]),
+    stubLogin: user => Promise.all([auth.stubLogin(user), elite2api.stubUser(user), elite2api.stubUserCaseloads()]),
```
andrewrlee, line 16: could you default this to `{}` here and avoid having to pass in empty object everywhere?

integration-tests/plugins/index.js
```diff
@@ -40,5 +40,7 @@ module.exports = on => {
     stubFindUsers: auth.stubFindUsers,
 
     stubUnverifiedUserDetailsRetrieval: auth.stubUnverifiedUserDetailsRetrieval,
+
+    stubVerifyToken: (active = true) => auth.stubVerifyToken(active),
```
andrewrlee, line 44: `stubVerifyToken: auth.stubVerifyToken,` ? so you don't need to do the defaulting in two places?

integration-tests/plugins/index.js
GurnankCheema: yes

integration-tests/plugins/index.js
GurnankCheema: correct

andrewrlee (approved):
LGTM

PR #449, opened 2021-02-11
DCS-805 pinning to minor version rather than patch

andrewrlee:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14-buster as builder
+FROM node:14.15-buster as builder
```
petergphillips, line 1: you missed the most important one - line 27 :grin:

petergphillips (dismissed):
LGTM

Dockerfile
andrewrlee: 😞  Done that now

mikehalmamoj (approved):

PR #450, opened 2021-02-24
DCS-863 Adding caseload and prison to add prison staff screens

andrewrlee:
List of selected users: 
<KBD>
Screenshot 2021-02-24 at 09 45 53
</KBD>

When more than one user:
<kbd>
Screenshot 2021-02-24 at 09 47 23
</kbd>

### DIFF
--- a/assets/sass/local.sass
+++ b/assets/sass/local.sass
@@ -387,19 +387,15 @@ dt.summary-list__key__wider
 .width-half
   width: 50% !important
 
-.added-involved-staff
-  th
-    border-bottom: none
-  td
-    border-bottom: none
-
-.email
+.found-staff-detail
   padding-top: 8px
+  padding-bottom: 17px !important
 
 .user-heading
   padding-left: 59px
 
-.staff-match-email
+.staff-match
+  padding-top: 8px
   cursor: pointer
   &:empty
     content: ""
@@ -417,6 +413,9 @@ dt.summary-list__key__wider
 .staff-involved-padding
   padding-inline-start: 0
   list-style: none
+
+.delete-staff
+  text-align: right
 
 .min-width-33
   min-width: 33%
--- a/server/data/authClientBuilder.ts
+++ b/server/data/authClientBuilder.ts
@@ -85,6 +85,7 @@ export type EmailResult = {
 
 type UserResult = {
   name: string
+  activeCaseLoadId?: string
   staffId: number
 }
 
@@ -93,6 +94,7 @@ export type FoundUserResult = {
   verified: boolean
   email?: string // only if verified
   name: string
+  activeCaseLoadId?: string // not present for new users or having active caseload removed
   staffId: number
 }
 
@@ -118,6 +120,10 @@ export class AuthClient {
     const path = `${apiUrl}/api/user/${username}`
     const body = await this.get({ path })
     return body
+  }
+
+  async getUsers(usernames: string[]): Promise<UserResult[]> {
+    return Promise.all(usernames.map(username => this.getUser(username)))
   }
 
   async findUsers(firstName: string, lastName: string): Promise<FoundUserResult[]> {
--- a/server/data/statementsClient.ts
+++ b/server/data/statementsClient.ts
@@ -9,7 +9,7 @@ import type {
   StatementUpdate,
 } from './statementsClientTypes'
 import { Status } from '../services/statementServiceTypes'
-import type { DraftInvolvedStaff } from '../services/report/draftReportService'
+import type { DraftInvolvedStaff } from '../services/report/draftInvolvedStaffService'
 import { StatementStatus } from '../config/types'
 
 type StatementCreationResult = { id: number; userId: string }
--- a/server/routes/testutils/appSetup.ts
+++ b/server/routes/testutils/appSetup.ts
@@ -59,7 +59,12 @@ export const coordinatorUser = {
   activeCaseLoadId: 'LEI',
 }
 
-export const appSetup = (route, userSupplier = (): any => user, isProduction = false): Express => {
+export const appSetup = (
+  route,
+  userSupplier = (): any => user,
+  isProduction = false,
+  flash = jest.fn().mockReturnValue([])
+): Express => {
   const app = express()
 
   const mockTransactionalClient = { query: jest.fn(), release: jest.fn() }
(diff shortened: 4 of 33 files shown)


assets/sass/local.sass
```diff
@@ -387,19 +387,15 @@ dt.summary-list__key__wider
...
-.added-involved-staff
-  th
-    border-bottom: none
-  td
-    border-bottom: none
-
-.email
+.found-staff-detail
   padding-top: 8px
+  padding-bottom: 17px !important
```
andrewrlee, line 392: As we're not following govuk standard designs this is required.
We were using standard spacing but items slowly get out of line when large numbers of results are returned

server/data/authClientBuilder.ts
```diff
@@ -120,6 +122,10 @@ export class AuthClient {
     return body
   }
 
+  async getUsers(usernames: string[]): Promise<UserResult[]> {
+    return Promise.all(usernames.map(username => this.getUser(username)))
+  }
```
andrewrlee, line 127: could potentially add a bulk endpoint but numbers of users to retrieve isn't likely to be very large

server/data/statementsClient.ts
```diff
@@ -9,7 +9,7 @@ import type {
   StatementUpdate,
 } from './statementsClientTypes'
 import { Status } from '../services/statementServiceTypes'
-import type { DraftInvolvedStaff } from '../services/report/draftReportService'
+import type { DraftInvolvedStaff } from '../services/report/draftInvolvedStaffService'
```
andrewrlee, line 12: data layer relying on service layer - need to move this stuff down or create a `model`

server/routes/testutils/appSetup.ts
```diff
@@ -90,7 +96,8 @@ export const appSetup = (route, userSupplier = (): any => user, isProduction = f
 export const appWithAllRoutes = (
   overrides: Partial<Services> = {},
   userSupplier = () => user,
-  isProduction?: boolean
+  isProduction?: boolean,
+  flash = jest.fn().mockReturnValue([])
```
andrewrlee, line 100: allow stubbing of flash content in tests

server/services/index.ts
```diff
@@ -54,9 +55,16 @@ const submitDraftReportService = new SubmitDraftReportService(
...
+const draftInvolvedStaffService = new DraftInvolvedStaffService(
+  authClientBuilder,
+  elite2ClientBuilder,
+  draftReportClient,
+  userService
+)
 
 const draftReportService = new DraftReportService(
   draftReportClient,
+  draftInvolvedStaffService,
```
andrewrlee, line 67: this has loads of dependencies - could do with looking at breaking this down a bit.

server/services/report/draftInvolvedStaffService.ts
```diff
@@ -0,0 +1,60 @@
...
+    const [user] = users
+    return { ...user, isReporter: true }
+  }
+
+  /**
+   * With the add to list pattern, a user will no longer be able to explicitly add themselves as involved staff.
+   * So we could just always add the user here. Unfortunately current draft reports could theoretically have added users.
+   * TODO: Add a migration to remove the current user if present on reports? Or check if even necessary?
+   */
+  public async getInvolvedStaff(token: string, username: string, bookingId: number): Promise<DraftInvolvedStaff[]> {
```
andrewrlee, line 32: think it would be worth looking to see if we can address this todo

server/services/report/draftReportService.ts
```diff
@@ -30,28 +34,25 @@ export default class DraftReportService {
     return this.draftReportClient.get(userId, bookingId)
   }
 
-  private async reporter(token: string, reporterUsername: string): Promise<DraftInvolvedStaff> {
```
andrewrlee, line 33: extracted this out into new service with new method

server/views/formPages/addingStaff/staff-involved.html
```diff
@@ -19,10 +19,14 @@ <h1 class="govuk-heading-xl mainHeading">{{ pageTitle }}</h1>
             text: item.name
             },
             {
-            text: item.email
+              text: item.prison | default('None', true)
```
andrewrlee, line 22: by default `default` only applies the default when the value is `undefined`. You need to pass true as the second param to use the default when the value is any falsy object. 

In this case it can be null so we need to pass the flasg.

server/services/userService.ts
```diff
@@ -84,4 +85,42 @@ export default class UserService {
       throw error
     }
   }
+
+  sortUsers = (agencyId: string) => (user1: UserWithPrison, user2: UserWithPrison): number => {
```
bell-pa, line 89: This doesn't sort users, it compares users.  userComparator perhaps?

server/services/report/draftInvolvedStaffService.ts
```diff
@@ -0,0 +1,60 @@
...
+    private readonly userService: UserService
+  ) {}
+
+  private async reporter(token: string, reporterUsername: string): Promise<DraftInvolvedStaff> {
+    const users = await this.userService.getUsers(token, [reporterUsername])
+    if (!users.length) {
+      throw new Error(`cannot load reporter user: ${reporterUsername}`)
+    }
+    const [user] = users
+    return { ...user, isReporter: true }
```
bell-pa, line 24: Don't think you need to add an 'isReporter' property here. The function name says it all. In fact the function just retrieves a user. See below.

server/services/report/draftInvolvedStaffService.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * With the add to list pattern, a user will no longer be able to explicitly add themselves as involved staff.
+   * So we could just always add the user here. Unfortunately current draft reports could theoretically have added users.
+   * TODO: Add a migration to remove the current user if present on reports? Or check if even necessary?
+   */
+  public async getInvolvedStaff(token: string, username: string, bookingId: number): Promise<DraftInvolvedStaff[]> {
+    const retrievedStaff = await this.draftReportClient.getInvolvedStaff(username, bookingId)
+    const staffWithoutReporter = retrievedStaff.filter(staff => staff.username !== username)
+    return [await this.reporter(token, username), ...staffWithoutReporter]
```
bell-pa, line 35: Add the isReporter property here?  This is the function that knows what kind of a user is being retrieved.

bell-pa (approved):
LGTM

PR #452, opened 2021-02-25
DCS-878 Make retrieval of locations more robust

bell-pa:
Handled in two ways:
* When searching for details of a location ask prison-api to include inactive locations in the response.
* If prison-api replies 'Not Found' to a location request handle this and carry on.  The locationService already returns empty objects for particular conditions so I've assumed that its clients know how to deal with a Location that isn't.  The only change then is to ensure that elite2ClientBuilder.getLocation handles a 404 response by returning an empty object.


andrewrlee (approved):
Bit weird the returning an empty object but your cypress test shows its handled 🤷 

LGTM

PR #457, opened 2021-03-02
DCS-848 Add question to record which manager authorised the use of force. 

bell-pa:
:bug: Make sure the authorising manager is displayed in read-only views.


GurnankCheema (approved):
LGTM

PR #461, opened 2021-03-04
DCS-848 Fixing issue around showing stale data

andrewrlee:
Once a field had been stored on the incident details page, if you then cleared out the value and clicked save, it would show the correct validation message but the old field value.

Moved to use the same pattern that we use on other form pages, which is:

If there is any flash then use that otherwise use values from the DB.
Previously it was showing the fields from the db and then overriding with flash entries which doesn't work when data is not present at all in flash

### DIFF
--- a/server/routes/creatingReports/incidentDetails.test.ts
+++ b/server/routes/creatingReports/incidentDetails.test.ts
@@ -15,9 +15,10 @@ const offenderService = new OffenderService(null) as jest.Mocked<OffenderService
 const locationService = new LocationService(null) as jest.Mocked<LocationService>
 
 let app
+const flash = jest.fn()
 
 beforeEach(() => {
-  app = appWithAllRoutes({ draftReportService, offenderService, locationService })
+  app = appWithAllRoutes({ draftReportService, offenderService, locationService }, undefined, false, flash)
   draftReportService.getCurrentDraft.mockResolvedValue({})
   offenderService.getOffenderDetails.mockResolvedValue({
     displayName: 'Bob Smith',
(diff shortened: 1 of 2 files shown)


server/routes/creatingReports/incidentDetails.test.ts
```diff
@@ -65,6 +68,42 @@ describe('GET /section/form', () => {
...
+    return request(app)
+      .get(`/report/1/incident-details`)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Eric Bloodaxe')
+        expect(locationService.getIncidentLocations).toBeCalledWith('user1-system-token', 'current-agency-id')
+      })
+  })
+
+  test('should preference flash state over persisted data', () => {
```
GurnankCheema, line 89: Rather than testing for something that isn't there, would it be more definitive to test for something that is?  
~ setting a value for 'authorisedBy' in flash i.e 
flash.mockReturnValue([ plannedUseOfForce: true, authorisedBy: 'some person' ])
expect(res.text).toContain('some person)

?

GurnankCheema (approved):
LGTM

server/routes/creatingReports/incidentDetails.test.ts
andrewrlee: You are totally right that it's not nice testing something that isn' there.

so this is a bit weird, if any value is on the key in flash it would have worked correctly - it only takes into account when the key is missing from flash.

this is because two objects were splatted on top of each other:
```
const a = { bob: true }
const b = { bob: false}

const c =  { ...a, ...b}  // => { bob: false}
```

This wouldn't work when we removed the `b` key: 
```
const a = { bob: true }
const b = { }

const c =  { ...a, ...b}  // => wanted '{}' got '{ bob: true}'
```
So had to change the implementation to not use splatting

The way I fixed this was write this failing test first and then fixing it to make the test pass.

GurnankCheema: cool

PR #468, opened 2021-03-16
DCS-846 Adding pages for recording use of force reasons

andrewrlee:
These pages are currently hooked up to anything, will add integration tests when added to part of the create report flow


louisenorris (approved):
LGTM

PR #470, opened 2021-03-18
DCS 850 change to relocation question

GurnankCheema:
Relocation and injuries page now includes additional options for Was the prisoner compliant question. 
When selecting 'other', the user is provided with a free form text box. 

Screenshot 2021-03-18 at 12 14 44

### DIFF
--- a/integration-tests/pages/createReport/relocationAndInjuriesPage.js
+++ b/integration-tests/pages/createReport/relocationAndInjuriesPage.js
@@ -3,26 +3,49 @@ import EvidencePage from './evidencePage'
 
 const relocationAndInjuriesPage = () =>
   page('Relocation and injuries', {
+    prisonerRelocation: () => cy.get('[name="prisonerRelocation"]'),
     prisonerCompliant: () => cy.get('[name="relocationCompliancy"]'),
     relocationType: () => cy.get('[name="relocationType"]'),
+    typeOfRelocation: () => cy.get('[name="typeOfRelocation"]'),
+    healthcareInvolved: () => cy.get('[name="healthcareInvolved"]'),
+    healthcarePractionerName: () => cy.get('[name="healthcarePractionerName"]'),
+    prisonerInjuries: () => cy.get('[name="prisonerInjuries"]'),
+    f213CompletedBy: () => cy.get('[name="f213CompletedBy"]'),
+    prisonerHospitalisation: () => cy.get('[name="prisonerHospitalisation"]'),
+    staffMedicalAttention: () => cy.get('[name="staffMedicalAttention"]'),
+    staffNeedingMedicalAttentionName0: () => cy.get('[name="staffNeedingMedicalAttention[0][name]"]'),
+    staffNeedingMedicalAttentionHospitalisation0: () =>
+      cy.get('[name="staffNeedingMedicalAttention[0][hospitalisation]"]'),
+    addAnother: () => cy.get('[data-qa-add-another-staff-needing-medical-attention = true]'),
+    staffNeedingMedicalAttentionName1: () => cy.get('[name="staffNeedingMedicalAttention[1][name]"]'),
+    staffNeedingMedicalAttentionHospitalisation1: () =>
+      cy.get('[name="staffNeedingMedicalAttention[1][hospitalisation]"]'),
+    staffNeedingMedicalAttentionName2: () => cy.get('[name="staffNeedingMedicalAttention[2][name]"]'),
+    staffNeedingMedicalAttention2: () => cy.get('[name="staffNeedingMedicalAttention[2][hospitalisation]"]'),
+    removeStaffNeedingMedicalAttention: () =>
+      cy.get('.add-another-staff-needing-medical-attention .add-another__remove-button'),
+
     fillForm() {
-      cy.get('[name="prisonerRelocation"]').select('Segregation unit')
+      this.prisonerRelocation().select('Segregation unit')
       this.prisonerCompliant().check('true')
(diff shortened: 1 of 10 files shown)


integration-tests/pages/createReport/relocationAndInjuriesPage.js
```diff
@@ -3,26 +3,49 @@ import EvidencePage from './evidencePage'
...
+    staffNeedingMedicalAttentionHospitalisation1: () =>
+      cy.get('[name="staffNeedingMedicalAttention[1][hospitalisation]"]'),
+    staffNeedingMedicalAttentionName2: () => cy.get('[name="staffNeedingMedicalAttention[2][name]"]'),
+    staffNeedingMedicalAttention2: () => cy.get('[name="staffNeedingMedicalAttention[2][hospitalisation]"]'),
+    removeStaffNeedingMedicalAttention: () =>
+      cy.get('.add-another-staff-needing-medical-attention .add-another__remove-button'),
+
     fillForm() {
-      cy.get('[name="prisonerRelocation"]').select('Segregation unit')
+      this.prisonerRelocation().select('Segregation unit')
```
andrewrlee, line 29: nice 👍

server/config/forms/relocationAndInjuriesValidation.test.js
```diff
@@ -139,6 +141,59 @@ describe("'complete' schema", () => {
...
+
+      expect(errors).toEqual([
+        {
+          href: '#relocationType',
+          text: 'Select the type of relocation',
+        },
+      ])
+      expect(formResponse.relocationCompliancy).toEqual(false)
+      expect(formResponse.relocationType).toBe(undefined)
+    })
```
andrewrlee, line 160: can we have a test for what happens when the post passes back both a type and an other

server/config/forms/relocationAndInjuriesForm.js
```diff
@@ -29,6 +29,12 @@ const completeSchema = joi.object({
     otherwise: joi.any().strip(),
   }),
 
+  typeOfRelocation: joi.when('relocationType', {
```
andrewrlee, line 32: bit confusing having `relocationType` and `typeOfRelocation`

Could the name of the new field be something like `otherRelocationDescription` or something?

server/config/types.ts
```diff
@@ -57,9 +57,11 @@ export const StatementStatus = toEnum({
 })
 
 export const RelocationType = toEnum({
-  SIDE: { value: 'SIDE', label: 'Side relocation' },
+  PRIMARY: { value: 'PRIMARY', label: 'Primary relocation' },
```
andrewrlee, line 60: can't rename this/remove SIDE as existing data will rely on it being present to pull back the description.

Might be worth have an active/inactive flag which doesn't get rendered when displaying the form but can be used by `toLabel`

server/services/reportSummary.ts
```diff
@@ -70,14 +70,22 @@ const createUseOfForceDetails = (
   }
 }
 
+const getRelocationType = relocationType => {
+  return relocationType ? ` - ${toLabel(RelocationType, relocationType).toLowerCase()}` : ''
+}
+
+const checkRelocationType = relocationAndInjuries => {
```
andrewrlee, line 77: maybe `getNonCompliancyReason` would be a better method name?

server/config/forms/relocationAndInjuriesForm.js
andrewrlee: `userSpecifiedRelocationType` ? 
...
something else?

server/views/formPages/incident/relocationAndInjuries.html
```diff
@@ -52,6 +49,12 @@ <h1 class="govuk-heading-xl mainHeading">{{ pageTitle }}</h1>
         errorMessage: errors | findError('relocationType'),
         options: data.types.RelocationType | list | extractAttr('value') | toOptions('value', 'label')
       },
+      followUpQuestionForTypeOfRelocation: {
```
andrewrlee, line 52: Better to have clear boundaries between non generic and generic code.

On the surface this seems to be a generic macro `radiosWithNestedRadiosWithNestedOtherFieldTextbox`
but the implementation details/macro options are specific to one specific use case.

andrewrlee: There's no harm extracting out a very specific macro if it makes the implementation simpler. 

It made sense to make the other macros generic as we were planning to reuse them but in this case it might produce a more maintainable solution if we hardcoded values in the implementation. 

We could look at making the macro generic if it turned out we had more than one `radiosWithNestedRadiosWithNestedOtherFieldTextbox`?

server/config/types.ts
louisenorris: Implemented as suggested.

server/config/forms/relocationAndInjuriesForm.js
louisenorris: Implemented as suggested - we went with userSpecifiedRelocationType.

server/services/reportSummary.ts
louisenorris: Implemented as suggested.

GurnankCheema: done

server/config/forms/relocationAndInjuriesValidation.test.js
louisenorris: Test added.

andrewrlee (approved):
LGTM, barring the follow up refactoring of `radiosWithNestedRadiosWithNestedOtherFieldTextbox` as discussed

PR #472, opened 2021-03-22
DCS-914 fixed relocation compliancy preselect bug

louisenorris:
Fixed preselect bug and renamed macro sections to be more relevant to the relocation compliancy and relocation type questions.


bell-pa (approved):
LGTM

PR #477, opened 2021-03-23
DCS-846 Adding selecting UoF reasons into main create report flow

andrewrlee:


server/routes/creatingReports/whyWasUoFApplied.ts
```diff
@@ -5,52 +5,79 @@ import { paths } from '../../config/incident'
...
   public view() {
     return async (req: Request, res: Response): Promise<void> => {
       const { bookingId } = req.params
-      const { reasons } = await this.draftReportService.getSelectedReasonsForUoF(req.user.username, Number(bookingId))
+      const { isComplete, reasons } = await this.draftReportService.getUoFReasonState(
+        req.user.username,
+        Number(bookingId)
+      )
+
+      const selectedReasons = req.flash('clearingOutReasons')?.length ? [] : reasons
```
andrewrlee, line 24: if there are some persisted reasons and the user deselects them all and hits submit it will trigger the validation  message. 

in this scenario we don't want to present the previously persisted values as the user has explicitly removed all of them

server/routes/creatingReports/whyWasUoFApplied.ts
```diff
@@ -5,52 +5,79 @@ import { paths } from '../../config/incident'
...
+        req.flash('clearingOutReasons', true)
         req.flash('errors', [{ href: '#reasons', text: 'Select the reasons why use of force was applied' }])
         return res.redirect(req.originalUrl)
       }
 
-      await this.draftReportService.process(res.locals.user, Number(bookingId), FORM, { reasons })
+      if (submitType === SubmitType.SAVE_AND_RETURN) {
+        await this.draftReportService.process(res.locals.user, Number(bookingId), FORM, { reasons })
+        return res.redirect(paths.reportUseOfForce(bookingId))
+      }
```
andrewrlee, line 48: With multiple reasons, when the report isn't complete and the user clicks save and return we will allow them to do so by leaving the primary reason unanswered. 

Once the report is complete we do not allow them saving it in a partially complete form - any actual persistent update is only done once they have selected a primary reason

server/routes/creatingReports/whyWasUoFApplied.ts
```diff
@@ -5,52 +5,79 @@ import { paths } from '../../config/incident'
...
 
-      const { reasons, primaryReason } = await this.draftReportService.getSelectedReasonsForUoF(
+      const { primaryReason, reasons } = await this.draftReportService.getUoFReasonState(
         req.user.username,
         Number(bookingId)
       )
-      const selectedReasons = Object.values(UofReasons).filter(({ value }) => reasons.includes(value))
+
+      const updatedReasons = req.flash('reasons')
+      const reasonsToUse = updatedReasons.length ? updatedReasons : reasons
```
andrewrlee, line 70: To cope with backwards navigation from the use of force details, we present the user the persisted values (as there will be no flash content).

This has the side effect of meaning that If a user refreshes the screen (clears out flash) then they will be presented the old persisted values again.

Seems like an ok trade off as users are probably more likely to use backwards navigation compared to refreshing

server/routes/creatingReports/whyWasUoFApplied.ts
```diff
@@ -5,52 +5,79 @@ import { paths } from '../../config/incident'
...
+
+      if (reasonsToUse.length === 1) {
+        return res.redirect(paths.whyWasUofApplied(bookingId))
```
andrewrlee, line 73: if they had previously persisted a single reason and then went to select multiple reasons and hit refresh - then the single persisted value would appear as the sole option for primary reason. 

This seems a bit weird as there is no other way of selecting a primary reason when you only have one option so to avoid it we just redirect back to the select your reasons screen.

server/types/express.d.ts
```diff
@@ -9,7 +9,9 @@ declare namespace Express {
     session: any
     logout: () => void
     csrfToken: () => string
-    flash: any
+    flash(): { [key: string]: any[] }
+    flash(type: string, message: any): number
+    flash(message: string): any[]
```
andrewrlee, line 14: copied this slightly dodgy type info from BVL

server/utils/nunjucksSetup.ts
```diff
@@ -118,7 +118,7 @@ export default function configureNunjucks(app: Express.Application): nunjucks.En
     }))
   })
 
-  njkEnv.addFilter('toPagination', (pageData: PageMetaData, query: any = {}) => {
+  njkEnv.addFilter('toPagination', (pageData: PageMetaData, query: Record<string, unknown> = {}) => {
```
andrewrlee, line 121: fixing some type errors as very close to the warning limit

louisenorris (approved):
LGTM

PR #479, opened 2021-03-24
DCS-906: two utility functions for hashing

GurnankCheema:


server/utils/utils.test.js
```diff
@@ -95,3 +103,16 @@ describe('parseDate', () => {
...
+describe('hash', () => {
+  it('should return hash', () => {
+    // result generated via https://approsto.com/sha-generator/
+    const result = stringToHash('abcABC')
+    expect(result).toEqual('XVnsPWdyGrzz4FzZhZQ1s2agjqxtfKqd3tNx1JC1tM8=')
+  })
+
+  it('should verify hash', () => {
+    const result = isHashedString('XVnsPWdyGrzz4FzZhZQ1s2agjqxtfKqd3tNx1JC1tM8=', 'abcABC')
+    expect(result).toBe(true)
```
andrewrlee, line 116: probably also want to test negative cases

server/utils/utils.js
```diff
@@ -42,6 +43,14 @@ const properCaseFullName = name =>
         .map(properCaseName)
         .join(' ')
 
+const stringToHash = inputString => {
+  return crypto.createHash('sha256').update(inputString).digest('base64')
+}
+
+const isHashedString = (hashValue, originalString) => {
+  return hashValue === crypto.createHash('sha256').update(originalString).digest('base64')
```
andrewrlee, line 51: might be better to use stringToHash? 

missing type signatures too on these.

server/utils/utils.js
```diff
@@ -42,6 +43,14 @@ const properCaseFullName = name =>
         .map(properCaseName)
         .join(' ')
 
+const stringToHash = inputString => {
+  return crypto.createHash('sha256').update(inputString).digest('base64')
```
andrewrlee, line 47: maybe force passing a salt here?

server/utils/utils.test.js
GurnankCheema: yes

louisenorris (dismissed):
LGTM

server/utils/utils.js
andrewrlee: or read the salt value from config here so the calling code doesn't need to pass it?

server/utils/utils.js
GurnankCheema: changed module from js to ts
above changes made

server/utils/utils.js
GurnankCheema: ok. Have passed salt in a separate. Have also changed method name to isHashOfString as is more descriptive

server/utils/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const stringToHash = (inputString: string, salt: string): string => {
+  return crypto.createHash('sha256').update(`${inputString}${salt}`).digest('base64')
+}
+
+export const isHashOfString = (hashValue: string, originalString: string, salt: string): boolean => {
+  return hashValue === stringToHash(originalString, salt)
+}
+
+module.exports = {
```
andrewrlee, line 59: think this can be deleted?  not used with typescript import/export

server/utils/utils.ts
```diff
@@ -0,0 +1,74 @@
+import R from 'ramda'
+import crypto from 'crypto'
+import moment, { Moment } from 'moment'
+
+export const isNilOrEmpty = R.either(R.isEmpty, R.isNil)
+export const getIn = R.path
+export const { equals } = R
+export const firstItem = R.head
+export const mergeWithRight = R.mergeDeepRight
```
andrewrlee, line 9: think we can get rid of `mergeWithRight` as not being used

andrewrlee: and `getIn`

server/utils/utils.ts
GurnankCheema: aha!

server/utils/utils.ts
GurnankCheema: good idea. I didn't check for their use across that app

andrewrlee (approved):
LGTM

PR #485, opened 2021-03-25
Notification service now includes encrypted url

GurnankCheema:
Emails to UoF involved staff will include a url with which they can use to request their removal from the involved staff list.
The url is suffixed with an encrypted component whose salt (HASH_SECRET) value for preprod and prod will be held in k8 secrets rather than helm config


helm_deploy/values-dev.yaml
```diff
@@ -23,4 +23,5 @@ env:
   EMAIL_LOCATION_URL: https://dev.use-of-force.service.justice.gov.uk
   TOKENVERIFICATION_API_URL: https://token-verification-api-dev.prison.service.justice.gov.uk
   TOKENVERIFICATION_API_ENABLED: true
+  HASH_SECRET: 'someHashSecret'
```
bell-pa, line 26: I think HASH_SECRET should be in _envs.tpl not here.
Perhaps something like
```
  - name: HASH_SECRET
    valueFrom:
      secretKeyRef:
        name: {{ template "app.name" . }}
        key: HASH_SECRET
```
Then the secret needs adding to the application's kubernetes secrets.
HASH_SECRET is a bit vague as a name. URL_SIGNING_SECRET perhaps?

andrewrlee: (my fault on the rubbish name!)

GurnankCheema: done

bell-pa (approved):
LGTM

helm_deploy/values-dev.yaml
GurnankCheema: done

PR #498, opened 2021-04-06
DCS-905 adding signature check on unauthenticated page

andrewrlee:


integration-tests/integration/statements/request-removal.spec.js
```diff
@@ -17,14 +18,21 @@ context('Request removal', () => {
...
+          },
         ],
       })
       .then(result => result.EMILY_JONES)
 
   it('A user can submit their statement removal request', () => {
     seedReport().then(statementId => {
-      const requestRemovalPage = RequestRemovalPage.visit(statementId)
+      cy.task('updateStatementId', { oldId: statementId, newId: 10 })
+      const requestRemovalPage = RequestRemovalPage.visit(10, 'mrsSjv9LqjClfvuNubYx4ACALTltJvqLmXYE0TKz6Xo=')
```
andrewrlee, line 35: Wanted to reuse hash functions but can't as browser crypto and server crypto have different impls.
This is just the hardcoded hash for statement 10 with the local dev hardcoded secret

server/config.js
```diff
@@ -58,7 +58,7 @@ module.exports = {
         OVERDUE: get('TEMPLATE_REPORTER_OVERDUE', '1cd6cd3f-7d45-4487-b029-c2a1270e6be8'),
       },
     },
-    urlSigningSecret: get('URL_SIGNING_SECRET', 'someUrlSigningSecret'),
+    urlSigningSecret: get('URL_SIGNING_SECRET', 'someUrlSigningSecret', requiredInProduction),
```
andrewrlee, line 61: added a required in production flag to ensure this is set

server/routes/unauthenticated/index.ts
```diff
@@ -16,6 +16,7 @@ export default function UnauthenticatedRoutes(services: Services): Router {
   router.post('/request-removal/:statementId', flash(), asyncMiddleware(removalRequest.submit))
 
   router.get('/removal-requested', removalRequest.viewConfirmation)
+  router.get('/already-removed', asyncMiddleware(removalRequest.viewAlreadyRemoved))
```
andrewrlee, line 19: new page for when the link refers to a deleted statement

server/routes/unauthenticated/requestRemoval.ts
```diff
@@ -12,34 +15,50 @@ export default class RemovalRequest {
...
-  view: RequestHandler = async (req, res) => {
+  view: RequestHandler = async (req, res, next) => {
     const statementId = extractStatementId(req)
-    const token = await this.systemToken()
-    const report = await this.reportService.getAnonReportSummary(token, statementId)
+    const { signature } = req.query
 
-    const errors = req.flash('errors')
+    if (!isHashOfString(signature?.toString() || '', statementId.toString(), config.email.urlSigningSecret)) {
+      return next(createHttpError(404, 'Not found'))
```
andrewrlee, line 23: if sig is invalid consider it a 404

server/routes/unauthenticated/requestRemoval.ts
```diff
@@ -12,34 +15,50 @@ export default class RemovalRequest {
...
+    }
+
+    const report = await this.reportService.getAnonReportSummary(await this.systemToken(), statementId)
 
-    return res.render(`pages/statement/request-removal.html`, { errors, report })
+    if (!report) {
+      return res.redirect(paths.alreadyRemoved())
```
andrewrlee, line 29: deleted report means already removed

server/routes/unauthenticated/requestRemoval.ts
```diff
@@ -12,34 +15,50 @@ export default class RemovalRequest {
...
   }
 
-  submit: RequestHandler = async (req, res) => {
+  submit: RequestHandler = async (req, res, next) => {
     const statementId = extractStatementId(req)
-    const { reason } = req.body
+    const { reason, signature } = req.body
+
+    if (!isHashOfString(signature?.toString() || '', statementId.toString(), config.email.urlSigningSecret)) {
+      return next(createHttpError(404, 'Not found'))
```
andrewrlee, line 40: also checking on post. Sig passed through as hidden field

andrewrlee: we don't explicitly prevent updating a deleted report... maybe we should?

server/services/notificationService.js
```diff
@@ -1,7 +1,7 @@
 const { NotifyClient } = require('notifications-node-client')
 const moment = require('moment')
 const { email } = require('../config')
-const { stringToHash } = require('../utils/utils')
+const { stringToHash } = require('../utils/hash')
```
andrewrlee, line 4: I moved this into separate module as part of the wrangling with int tests.. think its probably fine here

server/services/reportService.ts
```diff
@@ -60,6 +60,9 @@ export default class ReportService {
 
   async getAnonReportSummary(token: string, statementId: number): Promise<AnonReportSummaryWithPrison | undefined> {
     const report = await this.incidentClient.getAnonReportSummary(statementId)
+    if (!report) {
+      return undefined
```
andrewrlee, line 64: need to handle when report has been deleted - this through a type error before

server/utils/hash.ts
```diff
@@ -0,0 +1,10 @@
+import crypto from 'crypto'
+
+export const stringToHash = (inputString: string, salt: string): string => {
+  return crypto.createHmac('sha256', salt).update(inputString).digest('base64')
```
andrewrlee, line 4: this now creates a HMAC which is not as susceptible to length-extension attacks

integration-tests/integration/statements/request-removal.spec.js
bell-pa: Could you ask the server for the hash value instead? 
Looks like cy.task() can return a value: https://docs.cypress.io/api/commands/task#Yields

> The task plugin event handler can return a value or a promise. The command will fail if undefined is returned or if the promise is resolved with undefined. This helps catch typos or cases where the task event is not handled.

server/routes/unauthenticated/requestRemoval.ts
bell-pa: Could this be simplified? Instead of
`isHashOfString(signature?.toString() || '', statementId.toString(), config.email.urlSigningSecret)`
have something like
`isSignatureValidForStatementId(signature, statementId)`
and push the 'noise' into the function?

bell-pa (approved):
I suggested a couple of changes.
Otherwise: LGTM.

server/routes/unauthenticated/requestRemoval.ts
andrewrlee: done this.

integration-tests/integration/statements/request-removal.spec.js
andrewrlee: excellent suggestion!

bell-pa (approved):
LGTM

PR #503, opened 2021-04-14
DCS-948 Use removal request date as source of truth of removal status

andrewrlee:
We used to set the status but this made it tricky to workout how to determine whether people were requesting to remove a pending statement or submitted one

It turned out we needed to present statements based on their status before requesting removal

### DIFF
--- a/server/data/incidentClient.test.ts
+++ b/server/data/incidentClient.test.ts
@@ -64,7 +64,7 @@ test('getIncompleteReportsForReviewer', () => {
 
   const isRemovalRequested = `(select count(*) from "v_statement" s
                       where r.id = s.report_id
-                      and s.statement_status = $4) > 0`
+                      and s.removal_requested_date is not null) > 0`
 
   incidentClient.getIncompleteReportsForReviewer('agency-1')
 
@@ -80,12 +80,7 @@ test('getIncompleteReportsForReviewer', () => {
           where r.status = $1
           and   r.agency_id = $2
           order by r.incident_date`,
-    values: [
-      ReportStatus.SUBMITTED.value,
-      'agency-1',
-      StatementStatus.PENDING.value,
-      StatementStatus.REMOVAL_REQUESTED.value,
-    ],
+    values: [ReportStatus.SUBMITTED.value, 'agency-1', StatementStatus.PENDING.value],
   })
 })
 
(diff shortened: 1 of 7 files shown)


server/services/statementService.ts
```diff
@@ -69,8 +69,6 @@ export default class StatementService {
 
   async refuseRequest(statementId: number): Promise<void> {
     logger.info(`Removal request refused for statement with id: ${statementId}`)
-    const statement = await this.statementsClient.getStatementForReviewer(statementId)
-    const statusToSet = statement.submittedDate ? StatementStatus.SUBMITTED : StatementStatus.PENDING
```
andrewrlee, line 73: no longer need to work out what the previous statement was when refusing requests

server/data/incidentClient.test.ts
```diff
@@ -80,12 +80,7 @@ test('getIncompleteReportsForReviewer', () => {
           where r.status = $1
           and   r.agency_id = $2
           order by r.incident_date`,
-    values: [
-      ReportStatus.SUBMITTED.value,
-      'agency-1',
-      StatementStatus.PENDING.value,
-      StatementStatus.REMOVAL_REQUESTED.value,
-    ],
+    values: [ReportStatus.SUBMITTED.value, 'agency-1', StatementStatus.PENDING.value],
```
louisenorris, line 83: Just wondering, would the `StatementStatus.PENDING.value` be accessed here is there are only 2 attribute placeholders for each status and agency_id?

server/data/incidentClient.ts
```diff
@@ -95,12 +95,7 @@ export default class IncidentClient {
           where r.status = $1
           and   r.agency_id = $2
           order by r.incident_date`,
-      values: [
-        ReportStatus.SUBMITTED.value,
-        agencyId,
-        StatementStatus.PENDING.value,
-        StatementStatus.REMOVAL_REQUESTED.value,
-      ],
+      values: [ReportStatus.SUBMITTED.value, agencyId, StatementStatus.PENDING.value],
```
louisenorris, line 98: Just wondering, would the `StatementStatus.PENDING.value` be accessed here is there are only 2 attribute placeholders for each status and agency_id?

louisenorris (approved):
LGTM

server/data/incidentClient.ts
andrewrlee: as discussed

PR #511, opened 2021-04-22
DCS-947 email reminders sent only if removal has not been requested

GurnankCheema:
Bug fix.
Identified email reminders being sent even though a removal request had been made by involved staff. 
DB client sql changed so only returns results where there is no removal_requested_date


andrewrlee (approved):
LGTM

PR #512, opened 2021-04-23
DCS-958 coordinator will be redirected to view-statements page

GurnankCheema:
This ticket was about preventing the coordinator landing on the /vew-removal-request page if they had already decided to disallow an involved member's request to be removed from the report. 
The normal flow is /view-statements —> /view-removal-request —> /staff-member-not-removed

The /staff-member-not-removed page contains a hyperlink to return back to /view-statements  but if the coordinator chooses to use the browser back button instead, the browser will try to return to /view-removal-request. We want to prevent this. So in all cases the return from /staff-member-not-removed should be to  /view-statements

There are 2 data items in the DB to indicate a current removal request: the removal_requested_date and the removal_requested_reason. If a coordinator is on the /staff-member-not-removed page, then they are rejecting the users request. Hence the removal_requested_date and reason are nulled. 
Have just used the presence/absence of the removal_requested_date to determine the page to display if user uses the browser back button. 

There aren’t any new cypress integration tests because there is no way to mimic a browser back action

### DIFF
--- a/server/data/statementsClient.test.ts
+++ b/server/data/statementsClient.test.ts
@@ -295,11 +295,9 @@ test('getRemovalRequest', async () => {
   statementsClient.getRemovalRequest(1)
 
   expect(query).toBeCalledWith({
-    text: `select removal_requested_reason "removalRequestedReason"
-              , (select count(*) from "v_statement" s
-                      where s.id = $1
-                      and s.removal_requested_date is not null) > 0 "isRemovalRequested" 
-              from v_statement 
+    text: `select s.removal_requested_reason "removalRequestedReason"
+              , s.removal_requested_date is not null "isRemovalRequested"
+              from v_statement s 
               where id = $1`,
     values: [1],
   })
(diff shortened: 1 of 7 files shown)


server/data/statementsClient.test.ts
```diff
@@ -291,11 +291,15 @@ test('refuseStatementRemoval', async () => {
...
-test('getRemovalRequestedReasonByStatementId', async () => {
-  statementsClient.getRemovalRequestedReasonByStatementId(1)
+test('getRemovalRequest', async () => {
+  statementsClient.getRemovalRequest(1)
 
   expect(query).toBeCalledWith({
-    text: `select removal_requested_reason  "removalRequestedReason" from v_statement where id = $1`,
+    text: `select removal_requested_reason "removalRequestedReason"
+              , (select count(*) from "v_statement" s
+    where s.removal_requested_date is not null) > 0 "isRemovalRequested"
```
andrewrlee, line 300: this isn't quite right - what does this calculate?: 
`select count(*) from "v_statement" s where s.removal_requested_date is not null`

GurnankCheema: needs       where s.id = $1 
i.e     
const isRemovalRequested = `(select count(*) from "v_statement" s
                      where s.id = $1 
                      and s.removal_requested_date is not null) > 0`

andrewrlee (approved):
LGTM

PR #522, opened 2021-05-10
DCS-967 Adding audit log and recording report lifecycle events

andrewrlee:


migrations/20210508000000_form_db.js
```diff
@@ -0,0 +1,11 @@
+exports.up = knex =>
+  knex.schema.createTable('report_log', table => {
+    table.increments('id').primary('pk_audit')
+    table.timestamp('timestamp').notNullable().defaultTo(knex.fn.now())
+    table.string('username', 50).notNullable()
+    table.bigInteger('report_id').notNullable()
+    table.string('action', 50).notNullable()
+    table.jsonb('details').nullable()
+  })
```
andrewrlee, line 9: based on the licences version

server/data/draftReportClient.test.ts
```diff
@@ -76,21 +80,20 @@ test('updateAgencyId', () => {
...
             ,   submitted_date = $2
             ,   updated_date = now()
-          where r.user_id = $3
-          and r.booking_id = $4
-          and r.status = $5
-          and r.sequence_no = (select max(r2.sequence_no) from report r2 where r2.booking_id = r.booking_id and user_id = r.user_id)`,
-    values: [ReportStatus.SUBMITTED.value, date, 'user1', 1, ReportStatus.IN_PROGRESS.value],
+          where r.id = $3
+          and r.status = $4`,
+    values: [ReportStatus.SUBMITTED.value, date, 1, ReportStatus.IN_PROGRESS.value],
```
andrewrlee, line 94: simplified this as we already had access to the form id

server/routes/maintainingReports/admin.ts
```diff
@@ -44,7 +45,7 @@ export default class AdminRoutes {
 
     try {
       const updatedSection = JSON.parse(form)
-      await this.reviewService.update(res.locals.user, parseInt(reportId, 10), formName, updatedSection)
+      await this.reportService.update(res.locals.user, parseInt(reportId, 10), formName, updatedSection)
```
andrewrlee, line 48: move this from the review service to report service

server/services/involvedStaffService.ts
```diff
@@ -81,7 +81,14 @@ export class InvolvedStaffService {
 
       if (report.status === ReportStatus.COMPLETE.value) {
         logger.info(`There are now pending statements on : ${reportId}, moving from 'COMPLETE' to 'SUBMITTED'`)
-        await this.incidentClient.changeStatus(reportId, ReportStatus.COMPLETE, ReportStatus.SUBMITTED, client)
+        // TODO provide real username
+        await this.incidentClient.changeStatus(
```
andrewrlee, line 85: Still a bit unsure about this. 

These are automatic transitions that can happen as a side effect  when a staff member is added/removed to the report.

We could look at passing the staff member who triggered the action - though its not like a conscious action the user did

server/services/reportService.ts
```diff
@@ -81,6 +86,46 @@ export default class ReportService {
       throw new Error(`Report: '${reportId}' does not exist`)
     }
 
-    await this.incidentClient.deleteReport(reportId)
+    await this.incidentClient.deleteReport(username, reportId)
+  }
+
+  private getUpdatedReport(
```
andrewrlee, line 92: this has been moved over from the review service.. seemed a bit weird that a service that was all about the reading reports should be doing updates

server/data/reportLogClient.ts
```diff
@@ -0,0 +1,31 @@
+import { QueryPerformer } from './dataAccess/db'
+
+type Action =
```
bell-pa, line 3: enum?

andrewrlee: it's actually easier to consume as a string union - you don't need to actually reference the enum type in the client and you still get compile time safety and code completion

migrations/20210508000000_form_db.js
bell-pa: Think you might need some indexes on this table.  report_id? username? timestamp?

bell-pa (approved):
LGTM

PR #527, opened 2021-05-25
DCS-1001 Tidying up retrieving user details

andrewrlee:
Code supported retrieving multiple staff user records in one go but only every needed to retrieve a single user


server/services/drafts/draftInvolvedStaffService.ts
```diff
@@ -14,23 +14,17 @@ export class DraftInvolvedStaffService {
...
-   * With the add to list pattern, a user will no longer be able to explicitly add themselves as involved staff.
-   * So we could just always add the user here. Unfortunately current draft reports could theoretically have added users.
-   * TODO: Add a migration to remove the current user if present on reports? Or check if even necessary?
-   */
   public async getInvolvedStaff(token: string, username: string, bookingId: number): Promise<DraftInvolvedStaff[]> {
     const retrievedStaff = await this.draftReportClient.getInvolvedStaff(username, bookingId)
-    const staffWithoutReporter = retrievedStaff.filter(staff => staff.username !== username)
     const user = await this.reporter(token, username)
-    return [{ ...user, isReporter: true }, ...staffWithoutReporter]
+    return [{ ...user, isReporter: true }, ...retrievedStaff]
```
andrewrlee, line 27: now always just adding the reporter to the list by default. 

The add to list pattern prevents adding the current user explicitly. Looked in prod and there is only a single in progress report where the user has been added directly which I've now removed

server/services/userService.ts
```diff
@@ -36,39 +32,25 @@ export default class UserService {
...
-        .map(({ username, verified, email, name, staffId, activeCaseLoadId }) => ({
-          username,
-          verified,
-          email,
-          name,
-          staffId,
-          activeCaseLoadId,
-        }))
-
-      return results
```
andrewrlee, line 71: The main change of this pr was to kill the gnarly-ness above ^

bell-pa (approved):
LGTM

PR #532, opened 2021-06-10
DCS-1001-bugfix for username

GurnankCheema:
Application error if coordinator attempted to add an involved staff who was already in the report. Error only happened if the username was input in lower or mixed case rather than all uppercase.
Fix was to supply the uppercased username string when checking if user was already in the report.


andrewrlee (approved):
LGTM

PR #538, opened 2021-07-22
DCS-1094: code refactored to move filtering to DB query

GurnankCheema:
1. moved filtering out in-progress reports from the draft report service to the db client which also now returns the locationId rather than its parent object, to simplify code
2. renamed a test that suggested reports could be saved without an incident date. In fact the code was related
to updating an existing draft report. Reentering a date is not mandatory as it may be the user wants to change other values instead. Test description now reflects this.


andrewrlee (approved):
Nice, you can see how it simplifies the code!

LGTM!

PR #539, opened 2021-07-26
DCS-1029 implemented new support page

louisenorris:


JamieDraperUK (approved):
LGTM

server/views/partials/footer.html
```diff
@@ -0,0 +1,26 @@
+<footer class="govuk-footer " role="contentinfo">
```
andrewrlee, line 1: is there a reason why we cant use the govuk footer macro with this?

louisenorris: Was initially following the html format of header and layout already in place but seems like no reason not to use the nunjucks macro - implemented as suggested

server/views/help-reporting-use-of-force.html
```diff
@@ -0,0 +1,38 @@
...
+            
+        <h2 class="govuk-heading-m govuk-!-padding-10">Technical support</h2>
+          <p>For technical support, you should contact the helpdesk by calling {{supportTelephone}}, or {{supportExtension}} from inside an establishment.</p>    
+      </div>
+    </div>
+  </div>
+{% endblock %}
+
+{% block footer %}
+{{  govukFooter({}) }}
```
andrewrlee, line 37: meh... blocks/if/for etc should increase indentation

andrewrlee (approved):
LGTM

server/views/help-reporting-use-of-force.html
louisenorris: Ah - of course sorry

PR #547, opened 2021-08-05
DCS-1135 fix for feedback survey link

louisenorris:
If non prison users log into UoF and try to access the feedback survey link the link is missing hostname and current url path information as this was only set when a current prison user is present. Updated populate current user middleware to set these parameters whether a user is present or not as a short term fix. Maybe in the future UoF should not allow non prison users to access it.  
Screenshot 2021-08-05 at 10 43 05


andrewrlee (approved):
LGTM

PR #554, opened 2021-08-24
DCS-1156 incident report page results corrected

GurnankCheema:
The 'Date to' filter for completed reports wasn't returning reports with an incident date that fell on
the 'date to' date because the value passed to the reviewService was only up to 00:00:00 hours of that day.
Bug corrected by passing a value equal to the end of the day (23:59:59) to the reviewService.


server/routes/maintainingReports/reviewer.ts
```diff
@@ -37,10 +37,16 @@ export default class ReviewerRoutes {
       dateTo: dateTo ? parseDate(dateTo, 'DD/MM/YYYY') : null,
     })
 
+    const searchCriteria = { ...query }
+
+    if (searchCriteria.dateTo) {
+      searchCriteria.dateTo = parseDate(dateTo, 'DD/MM/YYYY').endOf('day')
```
andrewrlee, line 43: to avoid the additional conditional , the mutability and the duplicate parsing; could you just add a `.endOf('day')` on line 37?

GurnankCheema: That was my first thoughts too but calling .endOf on line 37 isn't possible because parseDate() can potentially return null , i.e when date is not valid.

```
export const parseDate = (val: string | unknown, format: string): Moment | null => {
  if (!val) {
    return null
  }
  const date = moment(val, format, true)
  return date.isValid() ? date : **null**
}
```

andrewrlee: can you use `?.` ? 

`?.endOf('day')`

GurnankCheema: yes. that will work.

andrewrlee (approved):
LGTM

PR #562, opened 2021-11-11
DSC-1310 adding ingress annotations

michalnawrockidj:


helm_deploy/use-of-force/values.yaml
```diff
@@ -3,3 +3,6 @@
 
 generic-prometheus-alerts:
   targetApplication: use-of-force
+
+  ingress:
+    external-dns.alpha.kubernetes.io/aws-weight: "100"
```
andrewrlee, line 8: Not sure this is right? 

Don't think it should be nested under `generic-prometheus-alerts`, needs to be inside an annotations section

michalnawrockidj: fixed

andrewrlee (approved):
LGTM

PR #571, opened 2022-01-14
DCS-1394: 🔧 Fix Google Tag Manager Environments and 🔐 Update dependencies for security scan

connormaglynn:


server/views/tagManager/headContent.html
```diff
@@ -7,17 +7,16 @@
...
     w[l] = w[l] || []
     w[l].push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' })
-    var f = d.getElementsByTagName(s)[0],
+    const f = d.getElementsByTagName(s)[0],
       j = d.createElement(s),
       dl = l != 'dataLayer' ? '&l=' + l : ''
     j.async = true
-    j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
-    var n=d.querySelector('[nonce]');
+    j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl + '{{ googleTagManagerEnvironment | safe }}';
```
connormaglynn, line 14: The environment needed to be a part of the src in order for Tag Manager to pick it up. Else everything just defaults to the prod version tag. Doesn't affect anything besides our ability to deploy different versions of tags to different environments, which is useful for testing.

server/views/tagManager/headContent.html
```diff
@@ -7,17 +7,16 @@
   ;(function (w, d, s, l, i) {
     w[l] = w[l] || []
     w[l].push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' })
-    var f = d.getElementsByTagName(s)[0],
+    const f = d.getElementsByTagName(s)[0],
```
andrewrlee, line 10: Checking analytics it looks like 7.15% of users use IE 11.

There is some support for `const` and this should be alright, just need to be wary that `const` is not supported in for-in and for-of loops

andrewrlee (approved):
LGTM

PR #574, opened 2022-02-02
DCS-1417 added hmp five wells IP addresses to allow list in preprod and prod

louisenorris:


helm_deploy/values-preprod.yaml
```diff
@@ -71,3 +71,5 @@ allow_list:
   moj-official-tgw-preprod: "51.149.251.0/24"
   crc-rrp: "62.253.83.37/32"
   crc-pp-wwm: "5.153.255.210/32"
+  five-wells-1: "20.49.214.199"
+  five-wells-2: "20.49.214.228"
```
andrewrlee, line 75: think there are a couple more values required

louisenorris: Cool, adding these now

andrewrlee (approved):
LGTM

PR #576, opened 2022-02-03
DCS-1434 Fix cross origin embedded policy after helmet upgrade

andrewrlee:
This was preventing jquery from being loaded from the CDN


louisenorris (approved):
LGTM

PR #596, opened 2022-06-14
DCS-1672 Update alert channel for refresh job

andrewrlee:
This isn't automated so I've manually applied the change and updating here for future reference


petergphillips (approved):
LGTM

PR #598, opened 2022-06-16
User location displayed as title case

davidthomas22:
Amended the location data format to user description rather than all upper case description. Changed in check your answers and report detail sections

### DIFF
--- a/server/routes/creatingReports/checkYourAnswers.ts
+++ b/server/routes/creatingReports/checkYourAnswers.ts
@@ -32,10 +32,7 @@ export default class CheckAnswerRoutes {
 
     const offenderDetail = await this.offenderService.getOffenderDetails(token, parseInt(bookingId, 10))
 
-    const { userDescription: locationDescription = '' } = await this.locationService.getLocation(
-      token,
-      form.incidentDetails.locationId
-    )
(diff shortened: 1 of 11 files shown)


server/routes/creatingReports/checkYourAnswers.ts
```diff
@@ -32,7 +32,7 @@ export default class CheckAnswerRoutes {
 
     const offenderDetail = await this.offenderService.getOffenderDetails(token, parseInt(bookingId, 10))
 
-    const { description: locationDescription = '' } = await this.locationService.getLocation(
+    const { userDescription: locationDescription = '' } = await this.locationService.getLocation(
```
andrewrlee, line 35: it looks like when that when we render locations for the list we do 
```
  userDescription: location.userDescription ? location.userDescription : location.locationPrefix,
``` 

I think we should be consistent and do the same for single locations.

It would be good to push this down into the service - we could make it a bit simpler on all the call sites if we just returned the description as a string , e.g:
```
 const locationDescription = await this.locationService.getLocation(token, form.incidentDetails.locationId)
```

GurnankCheema: yes, good point.

GurnankCheema (approved):
LGTM

PR #603, opened 2022-08-25
DCS-1816: ✨ Add dual running of ingress versions 0.47 and 1.2

connormaglynn:


helm_deploy/use-of-force/templates/ingress.yaml
```diff
@@ -10,7 +10,8 @@ metadata:
   annotations:
     kubernetes.io/ingress.class: "nginx"
     nginx.ingress.kubernetes.io/custom-http-errors: "418"
-    {{- toYaml .Values.ingress.annotations | nindent 4 }}
+    external-dns.alpha.kubernetes.io/set-identifier: use-of-force-{{ $fullName }}-green
```
connormaglynn, line 13: so using **use-of-force-{{ $fullName }}-green** seemed like the easiest way I could get the identifier to carry on being **use-of-force-use-of-force-${env}-green**

helm_deploy/use-of-force/templates/ingress_v1_2.yaml
```diff
@@ -0,0 +1,32 @@
...
+{{- $ingressPath := .Values.ingress.path -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ $ingressName }}
+  labels:
+    {{- include "app.labels" . | nindent 4 }}
+  annotations:
+    nginx.ingress.kubernetes.io/custom-http-errors: "418"
+    external-dns.alpha.kubernetes.io/set-identifier: {{ $ingressName }}-{{ .Release.Namespace }}-green
```
connormaglynn, line 13: This is what the generic-serivce is doing (seems slightly different to what we we're doing before). Though I thought it'd be better to align with the generic-service now whilst we're moving Ingress anyways.

Oddly when templated, it's showing up as **use-of-force-prod-v1-2-default-green**. Assuming this is because it's not running in the context of the CircleCI **deploy_env** job.

Thought it was worth mentioning in case anyone has better foresight than I do. I suppose we'll see when it gets deployed to dev if the names have stayed the same 👀

helm_deploy/use-of-force/values.yaml
```diff
@@ -11,5 +11,5 @@ generic-data-analytics-extractor:
   enabled: true
 
 ingress:
-  annotations:
```
connormaglynn, line 14: Thought I'd completely remove the annotations, since we're no longer using them

helm_deploy/use-of-force/values.yaml
```diff
@@ -11,5 +11,5 @@ generic-data-analytics-extractor:
...
-    external-dns.alpha.kubernetes.io/aws-weight: "100"
+  v1_2_enabled: true
+  v0_47_enabled: true
```
connormaglynn, line 15: 🤞 works like the generic-service implementation, where we should see the traffic slowly move over.

helm_deploy/use-of-force/templates/ingress.yaml
andrewrlee: meh.. guess you could have just hardcoded it

helm_deploy/use-of-force/templates/ingress_v1_2.yaml
andrewrlee: wonder if it might be because we have a default namespace set in the kubeconfig context and helm reads it from there? 
https://github.com/ministryofjustice/hmpps-circleci-orb/blob/e2abfc77c9bb08bdf8c96cf9e57c841f131b052f/src/commands/k8s_setup.yml#L33-L36

andrewrlee (approved):
LGTM

PR #609, opened 2022-10-11
DCS-1827: 🔥Remove old ingress ⬆️update dependencies

connormaglynn:


helm_deploy/use-of-force/values.yaml
```diff
@@ -12,4 +12,3 @@ generic-data-analytics-extractor:
 
 ingress:
   v1_2_enabled: true
-  v0_47_enabled: true
```
connormaglynn, line 15: Deleting this flag and deleted the ingress.yaml file. I thought I'd keep the new one named as v1_2 as it keeps it in par with the generic service. Hopefully makes it easier when / if we migrate to it.

package.json
```diff
@@ -79,7 +79,7 @@
       "test/*",
       "integration-tests/*"
     ],
-    "delay": "2500",
+    "delay": 2500,
```
connormaglynn, line 82: popped up in my ide that this should be a integer

.circleci/config.yml
```diff
@@ -1,8 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@3.12
-  slack: circleci/slack@4.4.2
+  hmpps: ministryofjustice/hmpps@6
```
connormaglynn, line 4: Big version jump. Not entirely sure about the ramifications of v4 and v5, but I know v6 is safe. I've seen v3 to v5 done in [this PR](https://github.com/ministryofjustice/manage-soc-cases/pull/331)so just chasing up on that in case there's anything to look out for

connormaglynn: Had a chat with John who did the manage-soc-cases PR. Says it was only the version bump from v3 to v5 that was required as part of the change.

andrewrlee (approved):
lgtm

PR #617, opened 2023-01-24
DCS-2017 patch and minor updates to dependencies

GurnankCheema:
The essential update was to superagent as it has a dependency on cookiejar which was identified in a security audit.


andrewrlee (approved):
LGTM

PR #626, opened 2023-04-03
DCS-2142 UoF revert service unavailable commits

GurnankCheema:
To remove all the changes associated with the service outage banner and page
This is to be deployed to prod AFTER the DB update has been done.


davidthomas22 (approved):
LGTM

PR #629, opened 2023-05-10
DCS-2153-update-postgres-job-config

davidthomas22:


database_refresh/02-cronjob.yaml
```diff
@@ -18,7 +18,7 @@
                 runAsUser: 999
               containers:
               - name: dbrefresh
-                image: "postgres:11.4"
+                image: "postgres:14.3"
```
andrewrlee, line 21: Maybe could update [docker-compose](https://github.com/ministryofjustice/use-of-force/blob/main/docker-compose.yml#L15) too?

andrewrlee (approved):
LGTM

PR #645, opened 2023-09-15
MAP-312 display service outage banner. 

GurnankCheema:
Required because Nomis will be down
FEATURE_FLAG_OUTAGE_BANNER_ENABLED to be set to true in kube deployment directly


davidthomas22 (approved):
LGTM

PR #650, opened 2023-10-05
MAP-305: Use HMPPS Manage Users API for user details

jimbali:
This has been split out into a new API. It was once part of HMPPS Auth.

https://dsdmoj.atlassian.net/browse/MAP-305

### DIFF
--- a/server/data/authClient.ts
+++ b/server/data/authClient.ts
@@ -13,7 +13,8 @@ const timeoutSpec = {
   response: config.apis.oauth2.timeout.response,
   deadline: config.apis.oauth2.timeout.deadline,
 }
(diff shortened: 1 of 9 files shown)


davidthomas22 (approved):
LGTM

server/data/authClient.ts
```diff
@@ -147,7 +148,7 @@ export class AuthClient {
   }
 
   async findUsers(firstName: string, lastName: string): Promise<FoundUserResult[]> {
-    const path = `${apiUrl}/api/prisonuser`
+    const path = `${hmppsManageUsersApiUrl}/prisonusers`
```
GurnankCheema, line 151: prisonuser**s** rather than user?

PR #666, opened 2024-01-16
MAP-577 New options for Prisoner Relocation

danbenton-mojdt:


GurnankCheema (approved):
LGTM

package.json
```diff
@@ -29,8 +29,8 @@
     "security_audit": "npx audit-ci --config audit-ci.json"
   },
   "engines": {
-    "node": "^18",
-    "npm": "<10"
+    "node": "^21",
```
jimbali, line 32: Do we definitely want to use Node 21? Only even numbers are LTS versions

jimbali (approved):

PR #671, opened 2024-01-18
MAP-573 Add escorting hold question

danbenton-mojdt:


server/data/UseOfForceReport.ts
```diff
@@ -14,6 +14,7 @@ export type UseOfForceDetails = {
   pavaUsed: boolean
   guidingHold: boolean
   guidingHoldOfficersInvolved: number
+  escortingHold?: boolean
```
GurnankCheema, line 17: is this optional ?

GurnankCheema (approved):
LGTM

server/data/UseOfForceReport.ts
danbenton-mojdt: I think so, given previously created reports won't have a value for it

PR #691, opened 2024-04-15
MAP-954 make node base same as typescript template project

GurnankCheema:


danbenton-mojdt (changes requested):
Circle config still on 20.11

danbenton-mojdt (approved):

PR #703, opened 2024-06-18
MAP-1280 Switch to global pem for certificate upgrade

danbenton-mojdt:


Mjwillis (approved):
LGTM - although there is a way not to install curl

PR #708, opened 2024-08-19
MAP-148: Remove cell location options for HMP Humber

ashleygyngell:


danbenton-mojdt (changes requested):
Nice work, could we call the env FEATURE_FLAG_REMOVE_CELL_LOCATION_AGENCIES (if not too long) and have it handle that value being a comma separated list of agencies, adding another test for having multiple agencies flagged?

server/services/locationService.test.ts
```diff
@@ -217,5 +224,29 @@ describe('locationService', () => {
 
       expect(prisonClientBuilder).toBeCalledWith(token)
     })
+
+    it('should remove cell location options for prisons that are in and out of the feature flag', async () => {
```
danbenton-mojdt, line 228: Nice

danbenton-mojdt (approved):

PR #712, opened 2024-09-16
Map 1548 uof use locations api to get location data

GurnankCheema:
[MAP-1548](https://dsdmoj.atlassian.net/jira/software/c/projects/MAP/boards/1354?selectedIssue=MAP-1548)
The code change in this ticket swaps the retrieval of the internal location data within a prison from prison-api to the new locations-inside-prison-api.

In prison-api, locations id's are an integer value but in locations-inside-prison-api the location id's are uuid's. 
This difference means the the id from locations-inside-prison-api can not be saved against the same key in the DB.
Therefore a new key of incidentLocationId has been introduced to hold the UUID form of the id. 

Incident location's are encountrered in 3 main places: 
    - /incident-details when making a brand new report 
    - /check-your-answers also when making a brand new report 
    - /view-report for existing reports

There is no complication in the case of a new report as the UUID will just be assigned to the new incidentLocationId key.
However for /view-report, the DB record will contain the nomis locationId. Because the new locations api does not store the nomis locationId, it won't be possible to obtain the location description from the location-api using it. 
So we will use the nomis-sync-prisoner-mapping service to look up the equivalent location id (called dpsLocationId) and use it when calling the locations-api to get the location description. 

The mapping of the main elements between the prison-api and locations-api location object is like this:

**prison-api                |          locations-api               |             example**
descripton                   |         pathHierarchy              |           'EDUC-EDUC'
userDescription          |         localName                     |          'Res 2 Education'
locationPrefix              |          key                                |         'MDI-EDUC-EDUC'

[MAP-1548]: https://dsdmoj.atlassian.net/browse/MAP-1548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Just for clarity:

- locationId is the (current) variable used to store prisonApi's location id 
- incidentLocationId is the (new) variable  used to store the locationsInsidePrisonApi's location id 
- nomisLocationId is the actual key in the object returned by locationsInsidePrisonApi and it relates to the prisonApis location id
- dpsLocationId is the actual key in the object returned by locationsInsidePrisonApi


.circleci/config.yml
```diff
@@ -96,7 +96,7 @@ jobs:
           command: sudo apt-get install libxss1
       - run:
           name: Get wiremock
-          command: curl -o wiremock.jar https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-standalone/2.27.1/wiremock-standalone-2.27.1.jar
```
GurnankCheema, line 99: wiremock updatd as per template project

integration-tests/integration/createIncident/view-your-report.cy.js
```diff
@@ -16,8 +16,46 @@ context('A reporter views their own report', () => {
     cy.task('stubUserDetailsRetrieval', ['MR_ZAGATO', 'MRS_JONES', 'TEST_USER'])
   })
 
+  it("A user's report should contain the incident location description even if DB record only contains the nomis format (i.e integer format) of the location's Id", () => {
```
GurnankCheema, line 19: this is verifying that even if the report only contains the prisonApi location id, as is the case for existing reports, the locations description will still be displayed at the /view-report endpoint

integration-tests/integration/reportPages/check-your-answers.cy.js
```diff
@@ -30,6 +30,12 @@ context('Check your answers page', () => {
     cy.login()
   })
```
GurnankCheema, line 32: this verifies that in the 'straight-forward' case of a new report being created and the locationsInsidePrisonApi's location id is persisted, the correct location description is displayed in /check-your-answers

server/config/forms/incidentDetailsForm.js
```diff
@@ -65,7 +59,9 @@ const requiredIncidentDate = joi
 const transientSchema = joi.object({
   incidentDate: requiredIncidentDate.meta({ fieldType: EXTRACTED }).alter(optionalForPartialValidation),
 
-  locationId: requiredIntegerMsg('Select the location of the incident').alter(optionalForPartialValidation),
```
GurnankCheema, line 68: Both incidentLocationId and locationId are needed in the validation schema because the /incident-details page encountered when making a new report requires  incidentLocationId to be selected. However in /check-your-answers, we need to ensure validation passes where a user may be on the page when this new code change is deployed and hence may only have the locationId persisted to db. Not having the locationId in the validation schema will cause validation to fail even though it previously passed and will cause the app to fail when user submits the form

server/routes/creatingReports/checkYourAnswers.ts
```diff
@@ -11,28 +12,46 @@ export default class CheckAnswerRoutes {
...
       form = {},
       incidentDate,
       agencyId: prisonId,
     } = await this.draftReportService.getCurrentDraft(req.user.username, parseInt(bookingId, 10))
+
+    // At this point the reportStatus may be 'incomplete' (because of the validation rules) if only the nomis locationId is present.
+    // This scenario would occur where a user is already at /check-your-answers when the new code to use incidentLocationId is deployed
+    // To pass validation we need the equivalent dpsLocationId persisted to the db as incidentLocationId.
+    // The following 'if-block' does that.
+
```
GurnankCheema, line 33: if at the /check-your-answers page only the prisonApi id exists, look up the equivalent locationsApi id and save it to db, This will ensure validation does not fail

Mjwillis (approved):
Nice work

server/config/types.ts
```diff
@@ -181,3 +181,14 @@ export const Destinations = {
   TASKLIST: 'back-to-task-list',
   CHECK_YOUR_ANSWERS: 'check-your-answers',
 }
+
```
GurnankCheema, line 184: types acceptable to locationsInsidePrisonApi

server/data/draftReportClient.ts
```diff
@@ -94,7 +94,7 @@ export default class DraftReportClient {
   async getDuplicateReports(bookingId: number, [startDate, endDate]: DateRange): Promise<OffenderReport[]> {
     const results = await this.query({
       text: `select r.incident_date date
```
GurnankCheema, line 96: now when making a new report it will be incidentLocationId that will be persisted (although locationId may also be persisted if the code change is deployed while a report is partly completed)

server/data/locationClient.ts
```diff
@@ -0,0 +1,34 @@
+import logger from '../../log'
+import type { RestClient } from './restClient'
+import type { LocationInPrison } from './locationClientTypes'
+import { NonResidentialUsageType } from '../config/types'
+
+export default class LocationClient {
+  constructor(private restClient: RestClient) {}
+
```
GurnankCheema, line 8: UUID must be 36 chars long including hyphens

server/data/nomisMappingClient.ts
```diff
@@ -0,0 +1,12 @@
+import logger from '../../log'
+import type { RestClient } from './restClient'
+import type { LocationMapping } from './nomisMappingClientTypes'
+
+export default class NomisMappingClient {
+  constructor(private restClient: RestClient) {}
+
+  async getDpsLocationMappingUsingNomisLocationId(nomisLocationId: number): Promise<LocationMapping> {
```
GurnankCheema, line 8: getting the location's id from the locationsInsidePrisonApi but using the Nomis version of the id

PR #722, opened 2024-10-10
MAP-1670: Restore the correct liveness and readiness probes

jimbali:
When we updated the Helm chart it brought some default liveness and readiness probe config that is incompatible with this project. We need to override the probe config in the same way that the TypeScript template project does.

MAP-1670


jnioche-jd:
see https://github.com/ministryofjustice/hmpps-template-typescript/blob/1ff6e4ca1f4960049d031023a12fe15768f8dca7/helm_deploy/hmpps-template-typescript/values.yaml#L20

jnioche-jd (approved):

PR #725, opened 2024-10-15
MAP-1625 enabling update of incidentDetails with incidentLocationId

GurnankCheema:
This code change is to allow users to add the incidentLocationId to existing reports. The trigger for the action is by calling /update-location-id followed by /1/9, where 1 is the reportId id to start from and 9 the reportId to end on.

The incidentLocationId is obtained from the nomis-sync-prisoner-mapping api. The only part of the report that will change is the form_response.incidentDetails.
 Any reports that already have the incidentLocationId will not change.

The actual update action in the database uses the already available functionality used when we edit reports.

### DIFF
--- a/server/routes/adhocReportActions/addhocActionController.ts
+++ b/server/routes/adhocReportActions/addhocActionController.ts
@@ -1,0 +1,53 @@
+/* eslint-disable no-await-in-loop */
+import { Request, Response } from 'express'
+import type ReportService from '../../services/reportService'
+import { SystemToken } from '../../types/uof'
+import NomisMappingService from '../../services/nomisMappingService'
+import logger from '../../../log'
+
+export default class AddhocActionController {
+  constructor(
+    private readonly reportService: ReportService,
+    private readonly nomisMappingService: NomisMappingService,
+    private readonly systemToken: SystemToken
+  ) {}
+
+  public updateReport = async (req: Request, res: Response): Promise<void> => {
+    const { fromReportId, toReportId } = req.params
+    const fromId = parseInt(fromReportId, 10)
+    const toId = parseInt(toReportId, 10)
+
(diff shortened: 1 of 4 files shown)


server/routes/adhocReportActions/addhocActionController.ts
```diff
@@ -0,0 +1,53 @@
...
+          logger.info(
+            `Ad-hoc exercise: Report with id ${reportId} already has form_response.incidentDetails.incidentLocationId set`
+          )
+        } else if (locationId) {
+          const token = await this.systemToken(req.user.username)
+
+          const { dpsLocationId } =
+            await this.nomisMappingService.getDpsLocationDetailsHavingCorrespondingNomisLocationId(token, locationId)
+
+          report.form.incidentDetails.incidentLocationId = dpsLocationId
```
GurnankCheema, line 36: L36 add the new location id to the incidentDetails part of the report. No other parts of incidentDetails will change

server/routes/adhocReportActions/addhocActionController.ts
```diff
@@ -0,0 +1,53 @@
...
+
+          const updatedSection = report.form.incidentDetails
+
+          await this.reportService.update(res.locals.user, reportId, 'incidentDetails', updatedSection)
```
GurnankCheema, line 40: L40 , the update function is existing functionality that is used for editing reports

server/routes/adhocReportActions/addhocActionController.ts
```diff
@@ -0,0 +1,53 @@
...
+
```
GurnankCheema, line 41: I have started all the logs with the word 'Ad-hoc' just to be able to target them more easily

jnioche-jd (approved):

PR #734, opened 2024-11-07
MAP-1798 Move security jobs from circleci to github actions

marcusaleman:
See https://dsdmoj.atlassian.net/browse/MAP-1779

Relies on a repo variable set in GitHub


GurnankCheema (approved):
LGTM

.github/workflows/security_trivy.yml
```diff
@@ -0,0 +1,12 @@
+name: Security trivy dependency check
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "19 6 * * MON-FRI" # Every weekday
+jobs:
+  security-kotlin-trivy-check:
```
jimbali, line 7: We probably shouldn't call this "kotlin". It's not a Kotlin project.

PR #737, opened 2024-11-25
MAP-1874 increase time for which a job will be live

GurnankCheema:
the send-reminders cron jobs are intermittently failing. 
Increasing the 'live' time of the job as it may enable the job to complete before it fails
Note a new job is started every 300 seconds


michalnawrockidj (approved):
LGTM

PR #742, opened 2024-12-09
Revert "MAP-1937 update helm chart for data extractor (#740)"

GurnankCheema:
This reverts commit 4d9bc7e916b9da55c3232ad213f62af5c41f0779.


marcusaleman (approved):
it is same version than before so LGTM
Approve

PR #743, opened 2024-12-10
Map 1745 UOF bug users unable to access incident details page

GurnankCheema:
[MAP-1745](https://dsdmoj.atlassian.net/jira/software/c/projects/MAP/boards/1354?assignee=5cd29c554326200dc971eab7&selectedIssue=MAP-1745)

This code change fixes the scenario where the /incident-details page will not load if an offender is currently in a transfer or has left prison. This is signified by the offenders agencyId being TRN or OUT. 
In such cases, loading of incident-details fails (and throws a 400 error) because TRN and OUT are not actual prisons and getting their internal locations, which are required by the select drop-down on the page, would return an error.

With the new code, when a user selects /incident-details, a check of the agencyId is done. If it's either TRN or OUT then we redirect to the change prison page. The /change-prison endpoint has been renamed to /prison-of-incident because the user isn't necessarily expecting to change prison of their own accord. 

In the previous code, selecting the 'Change' link in /incident-details page saves any user input and calls draftReportService.process which either updates an existing report or creates a new one. This is triggered via frontend javascript that acts on components with the 'js-submitLink' html class. Because the frontend javascript trigger is now being circumvented by the immediate redirect to /prison-of-incident, the draftReportService.process is called in the submit of /prison-of-incident.

Have used flash to pass any user inputs to /prison-of-incident

[MAP-1745]: https://dsdmoj.atlassian.net/browse/MAP-1745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/routes/creatingReports/changePrison.test.ts
+++ b/server/routes/creatingReports/changePrison.test.ts
@@ -34,10 +34,10 @@ afterEach(() => {
   jest.resetAllMocks()
 })
 
-describe('GET /change-prison', () => {
+describe('GET /prison-of-incident', () => {
   test('should render content', () => {
     return request(app)
-      .get(`/report/-19/change-prison`)
+      .get(`/report/-19/prison-of-incident`)
       .expect('Content-Type', /html/)
       .expect(res => {
         expect(res.text).toContain('What prison did the use of force take place in?')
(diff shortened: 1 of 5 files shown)


BarryGee-MOJ (approved):
LGTM

PR #746, opened 2024-12-16
Map 1937 uof cron bug revert

GurnankCheema:
This reverts the changes in previous to commits because it has not made any changes to the cron job failing. 
Kubernetes is still killing the pod with reason _'Pod was active on the node longer than the specified deadline'_
Will investigate further but in meantime will revert ineffectual changes.


BarryGee-MOJ (approved):
LGTM

PR #756, opened 2025-02-19
MAP-2010 update dependencies

GurnankCheema:
1. non-breaking dependency update 
2. replaced csrf library and aligned it's implementation with template project


BarryGee-MOJ (approved):
lgtm

PR #757, opened 2025-02-20
MAP-2010 remove calls to old csrf library as now using csrf-sync

GurnankCheema:
The call to setUpCsrf() ensures csrf token is available for all endpoints as implemented in the template project. 
This includes the unauthenticated endpoints: /removal-requested, /already-removed and /removal-already-requested. 
These unauthenticated endpoints enable staff to be removed from the 'Staff involved' list via the email prompting them to create a statement. 
And even though these endpoints don't specifically set _csrf token in the html, it will not have an adverse affect because the html doesn't include any forms.
Note: the call to setUpCsrf() comes before the position of the now removed csrf() in the middleware call stack so will capture all endpoints


ashleygyngell (approved):
LGTM

PR #776, opened 2025-05-19
MAP-2319 add feature flag 

GurnankCheema:
enabling edit features to be displayed in dev only


server/utils/nunjucksSetup.ts
```diff
@@ -40,6 +40,7 @@ export default function configureNunjucks(app: Express.Application): nunjucks.En
   njkEnv.addGlobal('apiClientId', config.apis.oauth2.apiClientId)
   njkEnv.addGlobal('featureFlagOutageBannerEnabled', config.featureFlagOutageBannerEnabled)
   njkEnv.addGlobal('digitalPrisonServiceUrl', config.apis.digitalPrisonServiceUrl)
+  njkEnv.addGlobal('featureFlagReportEditingEnabled', config.featureFlagReportEditingEnabled)
```
GurnankCheema, line 44: this makes the config global so you can access in any template like 
  {% if featureFlagReportEditingEnabled %}
      ... do stuff
  {%endif%}

BarryGee-MOJ (approved):
LGTM

PR #777, opened 2025-05-19
MAP-2319 db migration to better differentiate report owner edits.

GurnankCheema:
Also rename columns and add index on report_id


BarryGee-MOJ (approved):
LGTM

PR #783, opened 2025-05-22
MAP-808 Change instantiation of api clients and service classes

danbenton-mojdt:
Also fix incident macros that were broken by govuk dependcy update


GurnankCheema (approved):
LGTM

PR #784, opened 2025-05-27
Map 2319 uof redesign report pages to include tab

GurnankCheema:
Notes: 
Variable renderReportView  in the views is used to differentiate check-your-answers from view-report and your-report. All three use the same underlying macro but the display of check-your-answers is not to be changed.

Some integration tests have been skipped using 'xit'. This is deliberate and the functionality to remove/add nvolved staff and to delete reports is to be in reintroduced at a later stage and the skipped tests should serve as a reminder of the things to test for.

### DIFF
--- a/server/views/pages/reportDetailMacro.njk
+++ b/server/views/pages/reportDetailMacro.njk
@@ -1,4 +1,3 @@
-
 {% import "./reportDetailSectionsMacros.njk" as reportDetailsMacros %} 
 
 {% macro fieldValue(name, value, dataQa) %}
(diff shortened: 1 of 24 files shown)


server/views/pages/reportDetailMacro.njk
```diff
@@ -53,14 +52,75 @@
   </div>
 {% endmacro %}
 
-{% macro detail(data, bookingId, incidentId, user, print) %}
+{% macro detail(data, bookingId, incidentId, user, print, renderReportView) %}
+    {% if renderReportView %}
```
GurnankCheema, line 56: See note 1 in comments

PR #828, opened 2025-08-12
MAP-2636 clear flash before saving new values

GurnankCheema:
clearing out flash before reusing because it was causing unexpected results


dafhoustonmoj (approved):
LGTM

PR #833, opened 2025-09-02
Map 2636 uof edit incident details persist edits to edits table

GurnankCheema:
This PR is to enable edits to the incident-details page
Have added a DB migration so that changes will now be stored in a jsonb blob rather than individual columns (old_value_primary, old_value_secondary etc)
Use of the bob follows the pattern in the actual original report


BarryGee-MOJ (commented):
just one tiny linting fix.

BarryGee-MOJ (approved):
LGTM

PR #841, opened 2025-09-15
MAP-2781 reason for change page

GurnankCheema:
- make additional info text mandatory 
- correct error messages
- ensure any previously selected radios and entered text are redisplayed following error message alerts

### DIFF
--- a/server/config/edit/reasonForChangeErrorMessageConfig.ts
+++ b/server/config/edit/reasonForChangeErrorMessageConfig.ts
@@ -1,0 +1,18 @@
+export default {
+  useOfForceDetails: {
+    reason: 'Provide a reason for changing the use of force details',
+    anotherReason: 'Specify the reason for changing the use of force details',
+  },
+  incidentDetails: {
+    reason: 'Provide a reason for changing the incident details',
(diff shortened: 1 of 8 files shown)


server/config/edit/reasonForChangeErrorMessageConfig.ts
```diff
@@ -0,0 +1,18 @@
+export default {
```
GurnankCheema, line 1: put error messages in config

server/routes/maintainingReports/coordinator.ts
```diff
@@ -393,7 +404,7 @@ export default class CoordinatorRoutes {
         reportSection,
         changes,
         reason,
-        reasonText,
```
GurnankCheema, line 396: only persist text if reason is still anotherReasonForEdit

server/views/pages/viewIncident/edit-history.njk
```diff
@@ -70,17 +70,14 @@
             <tr>
               <td colspan="6"></td>
               <td>
```
GurnankCheema, line 72: additional information text no longer optional

PR #847, opened 2025-09-30
MAP-2487 Adding / Editing staff (co-ordinator role)

BarryGee-MOJ:


server/config/forms/reasonForAddingStaffForm.js
```diff
@@ -0,0 +1,32 @@
+const { joi } = require('./validations')
```
GurnankCheema: nice

server/routes/maintainingReports/coordinator.ts
```diff
@@ -877,11 +878,277 @@ export default class CoordinatorRoutes {
     return res.render('pages/coordinator/staff-member-not-removed.html', { data })
   }
 
-  viewAddInvolvedStaff: RequestHandler = async (req, res) => {
+  viewInvolvedStaff: RequestHandler = async (req, res) => {
```
GurnankCheema: could use extractReportId(req) instead of  const { reportId } = req.params, see L29

server/routes/maintainingReports/coordinator.ts
```diff
@@ -877,11 +878,277 @@ export default class CoordinatorRoutes {
...
+      // Re-perform search for pagination
+      userSearchResults = await this.involvedStaffService.findInvolvedStaffFuzzySearch(
+        await this.authService.getSystemClientToken(res.locals.user.username),
+        parseInt(reportId, 10),
+        username,
+        page
+      )
+      content = userSearchResults?.content || []
+    }
+
```
GurnankCheema: would this make L940 - L959 it clearer? 
```
await Promise.all(
  content.map(async (staffMember) => {
    try {

      const isExisting = await this.involvedStaffService.loadInvolvedStaffByUsername(
        reportId,
        staffMember.username
      )

      if (isExisting) {
        staffMember.isExistingInvolvedStaffMember = true
      }
    } catch (error) {
      log.error(
        `Error checking involved staff member "${staffMember.username}" for report ${reportId}: ${error?.message}`,
        error
      )
    }
  })
)

```

server/routes/maintainingReports/coordinator.ts
BarryGee-MOJ: Agreed, updated.

PR #851, opened 2025-10-02
MAP-2798 handle report edits where reasonsForUseOfForce does not exist

GurnankCheema:
Some old reports don't have reasonsForUseOfForce in their dataset. 
This code is to prevent an application crash for those reports by checking whether the data exists before trying to access the reasons value
see `report.form.reasonsForUseOfForce?.reasons || []` and `report.form.reasonsForUseOfForce?.primaryReason || ''`


dafhoustonmoj (approved):
LGTM

PR #997, opened 2026-05-07
Mapb 394 uof bugfix navigating back after edit

GurnankCheema:
This fixes an issue where users could edit a submitted report, complete the whole edit journey, and then use the browser back button to revisit earlier pages in the flow. Navigating backwards and forwards in this way could result in incomplete data being saved, particularly within the “use of force details” section of the report.

Most report sections follow a simple flow consisting of a single edit page, followed by
- /{reportId}/edit-report/reason-for-change
- /{reportId}/view-incident?tab=report

However, the “use of force details” section uses a multi-page journey:
- /{reportId}/edit-report/why-was-uof-applied
- /{reportId}/edit-report/what-was-the-primary-reason-of-uof (shown when more than one reason exists)
- /{reportId}/edit-report/use-of-force-details

followed by:
- /{reportId}/edit-report/reason-for-change
- /{reportId}/view-incident?tab=report

Edit data is stored in session state. The session is created when the first page in the edit flow is completed and is cleared once the user successfully  'Saves' on /{reportId}/edit-report/reason-for-change.

If users are able to navigate backwards from /{reportId}/view-incident?tab=report to:
- /{reportId}/edit-report/reason-for-change
- /{reportId}/edit-report/use-of-force-details
- /{reportId}/edit-report/what-was-the-primary-reason-of-uof

without returning to the start of the journey (/{reportId}/edit-report/why-was-uof-applied), then any data normally captured within that page will be missed resulting in incomplete data being saved to the report.

This fix prevents users from accessing the following pages via reverse browser navigation:
- /{reportId}/edit-report/reason-for-change
- /{reportId}/edit-report/use-of-force-details
- /{reportId}/edit-report/what-was-the-primary-reason-of-uof

To edit the “use of force details” section, users must now restart the journey from /{reportId}/edit-report/why-was-uof-applied

This issue does not adversely affect other edit journeys that consist of only a single page before /{reportId}/edit-report/reason-for-change because validation will prevent user from progressing without making a change.

### DIFF
--- a/integration-tests/integration/coordinator/edit-report/use-of-force-details.cy.js
+++ b/integration-tests/integration/coordinator/edit-report/use-of-force-details.cy.js
@@ -441,4 +441,70 @@ context("A use of force coordinator needs to edit a submitted report's Use of Fo
       reasonForChangePage.tableRowAndColHeading(row, 'new-value').should('contain', newValue)
     })
   })
+
+  it('can not navigate back using browser once edit has been submitted', () => {
+    const notCompletedIncidentsPage = NotCompletedIncidentsPage.goTo()
+    notCompletedIncidentsPage.viewIncidentLink().click()
+
+    const viewIncidentPage = ViewIncidentPage.verifyOnPage()
+    viewIncidentPage.editReportButton().click()
+
+    const editReportPage = EditReportPage.verifyOnPage()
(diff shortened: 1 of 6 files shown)


dafhoustonmoj (approved):
LGTM

PR #1030, opened 2026-06-19
MAPB-528 prevent user backtrack

GurnankCheema:
If a coordinator rejects a removal request submitted by an involved staff member and reaches the `/staff-member-not-removed` page, they should not be able to return to the `/view-removal-request` page, as the refusal action has already been committed and is irreversible.

While the application already enforces this no-backtracking rule, it currently redirects users to the deprecated `/{reportId}/view-statements` page. The updated implementation correctly redirects users to the current `view-incident?tab=statements` page instead.


server/routes/maintainingReports/coordinator.ts
```diff
@@ -1215,7 +1216,7 @@ export default class CoordinatorRoutes {
     const errors = req.flash('errors')
 
     if (!removalRequest.isRemovalRequested) {
```
GurnankCheema, line 1218: the key change

dafhoustonmoj (approved):

